PR #43, opened 2020-10-13
Make data format dependencies optional

### LINKED ISSUE #40: Optional dependencies?
I'm not planning on using msgpack/toml/yaml in some small project, but they are always installed because pyserde requires them. As far as I understand the code, their use is isolated in separate submodules, so these dependencies can be safely uninstalled while related submodules are not used. So I suggest making these dependencies optional.

It might look something like this:

```python
requires = [...]
msgpack_requires = ['msgpack']
toml_requires = ['toml']
yaml_requires = ['pyyaml']
# ...
setup(
    # ...
    extras_require={
        'msgpack': msgpack_requires,
        'toml': toml_requires,
        'yaml': yaml_requires,
        'full': msgpack_requires + toml_requires + yaml_requires,
        'test': tests_require,
    },
    # ...
)
```

Then all dependencies can be installed with something like `pip install pyserde[full]`

andreymal:
Resolves #40

There are some things to keep in mind:

1. This is probably a breaking change; existing pyserde users may expect that all formats will be installed automatically (although if there are just a few users yet, this can be ignored..?)

2. If I understand correctly, pipenv is only used for development purposes which implies installing of all dependencies, so I didn't touch Pipfile.

3. Not sure what is the best way to describe the new installation process... I updated README.md according to my knowledge of English, but maybe there are better ways.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,18 @@ Install pyserde from PyPI. pyserde requires Python>=3.6.
 pip install pyserde
 ```
 
+Additional data formats besides JSON need additional dependencies installed. Install `msgpack`, `toml`, or `yaml` extras to work with the appropriate data formats; you can skip formats that you don't plan to use. For example, if you want to use Toml and YAML:
+
+```sh
+pip install pyserde[toml,yaml]
+```
+
+Or all at once:
+
+```sh
+pip install pyserde[all]
+```
+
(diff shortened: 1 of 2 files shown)


yukinarit:
Verified. Thank you!

```
$ pip install git+https://github.com/andreymal/pyserde.git@optional-deps#egg=pyserde[toml]

...

$ pip list
Package           Version
----------------- -------
Jinja2            2.11.2
MarkupSafe        1.1.1
mypy-extensions   0.4.3
pip               20.1.1
pyserde           0.1.5
setuptools        46.4.0
stringcase        1.2.0
toml              0.10.1
typing-extensions 3.7.4.3
typing-inspect    0.6.0
wheel             0.34.2
```

PR #55, opened 2020-11-01
Added example of a field with default_factory property

pranavvp10:
Added example of a field with default_factory property.

### DIFF
--- a/examples/default.py
+++ b/examples/default.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass, field
 from serde import deserialize, serialize
 from serde.json import from_json, to_json
+from typing import Dict
 
 
 @deserialize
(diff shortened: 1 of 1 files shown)


examples/default.py
```diff
@@ -11,13 +12,14 @@ class Foo:
...
     b: bool = field(default=True)
+    d: Dict[str,int] =field(default_factory=dict)
 
 
 def main():
     h = Foo()
     print(f"Into Json: {to_json(h)}")
 
-    s = '{"i": 10, "s": "foo", "f": 100.0, "b": true}'
+    s = '{"i": 10, "s": "foo", "f": 100.0, "b": true, "d":{}}'
```
yukinarit, line 22: Oops, sorry. 

Your code looks good to me, but can you please try to fix this line as follows?

```python
s = '{}'
```

you should still see the output like this.
```
From Json: Foo(i=10, s='foo', f=100.0, b=True, d={})
```

Because this is an example of `default` and `default_factory` field, `s = '{}'` makes `Foo` class uses the default/default_factory during deserialization. So I thought it's more appropriate 🙂

pranavvp10: Ok . I will do it

yukinarit: LGTM! Thank you! 👍

PR #68, opened 2020-11-16
CI improvements

### LINKED ISSUE #67: Installed python might not be used in CI
As you can see in the console, it's running in python 3.8 although it's during the `Python 3.7` stage.

Screen Shot 2020-11-16 at 18 54 50

It appears to me
* For `Ubuntu`: 
* For `macOS`: python3.8 is used during 3.6 / 3.7 / pypy3 stages
* For `Windows`: 
    * python 3.9 is used during 3.6 / 3.7 / 3.8 stages
    * python 3.7 is used during pypy stage

https://github.com/yukinarit/pyserde/runs/1404632126?check_suite_focus=true

@alexmisk Any idea? 🤔

alexmisk:
### 1. Explicitly pass Python version to pipenv (fix #67)
It looks like `pipenv` doesn't use default Python version on Windows and MacOS so I've decided to unfold your `make setup` statement for explicitly passing correct Python version to `pipenv`. You can check GitHub Action logs here https://github.com/alexmisk/pyserde/actions/runs/366167153 if needed.

This is still a workaround though. GitHub Actions workers are virtual environments *per se*, and using `pipenv` for creating another environment inside them seems like an overkill :-) One of the major downsides of `pipenv` inside CI is inability to deal with dependency caching effectively (which slows down CI workflow execution time).

### 2. Rename testing step for better readability
ci
Now one can see Python version and OS at the same time :-)

### 3. Change Python 3.9-dev to 3.9 in test matrix
Python 3.9 was released about a month ago, no need to use dev version anymore.

---
**P.S.** Note that benchmark step is still failing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,10 +19,15 @@ jobs:
        uses: actions/setup-python@v2
        with:
          python-version: 3.7
+     - name: Restore pip cache
+       uses: actions/cache@v2
+       with:
+         path: ${{ env.pythonLocation }}
+         key: ubuntu-latest-${{ env.pythonLocation }}-${{ hashFiles('**/Pipfile') }}
      - name: Install dependencies
        run: |
          pip install pipenv
-         make setup
+         pipenv install --skip-lock --system --dev
      - name: Get PR number
        uses: jwalton/gh-find-current-pr@v1
        id: findPr
@@ -39,7 +44,10 @@ jobs:
        env:
          PR: ${{ steps.findPr.outputs.pr }}
      - name: Build docs
-       run: make docs
+       run: |
+         mkdir -p docs
+         pdoc serde --html -o html --force
+         cp -f html/serde/* docs/
      - name: Deploy on gh-pages
        uses: JamesIves/github-pages-deploy-action@3.7.1
        with:
@@ -48,4 +56,3 @@ jobs:
          FOLDER: docs
          COMMIT_MESSAGE: ${{ env.COMMIT_MESSAGE }}
          CLEAN: true
-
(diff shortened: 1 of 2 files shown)


yukinarit:
LGTM and agree to the point you made.

>  GitHub Actions workers are virtual environments per se , and using pipenv for creating another environment inside them seems like an overkill :-) One of the major downsides of pipenv inside CI is inability to deal with dependency caching effectively (which slows down CI workflow execution time).

Maybe pipenv's --system option is meant for CI use case?

alexmisk:
> Maybe pipenv's --system option is meant for CI use case?

Looks like it, but I need more time to dig into the problem. I will work on pipenv/caching task during the week.

alexmisk:
Dependency caching in CI! Not brilliant yet, but quite tolerable.

### TL;DR

- up to 3x boost on docs building workflow
- up to 2x boost on test workflow
- cache is refreshing automatically when Pipfile is changed

---

This block caches the whole Python-related stuff:

```yaml
# build.yml
- name: Restore pip cache
       uses: actions/cache@v2
       with:
         path: ${{ env.pythonLocation }}
         key: ubuntu-latest-${{ env.pythonLocation }}-${{ hashFiles('**/Pipfile') }}
```

No caching on Windows in test matrix for now: Python misbehaves sometimes with restored dependencies.

```yaml
# test.yml
- name: Restore pip cache
      # Caching on Windows is turned off due to inconsistent behaviour
      if: matrix.os != 'windows-latest'
      uses: actions/cache@v2
      with:
        path: ${{ env.pythonLocation }}
        key: ${{ matrix.os }}-${{ env.pythonLocation }}-${{ hashFiles('**/Pipfile') }}
```

`pipenv --system` indeed installs dependencies system-wide and proper Python version is used without passing `--python ${{ matrix.python-version }}`

I've had to unfold all `make ...` statements to use default `python` and `pip`. `pipenv` always creates virtual environment when you try to run something.

yukinarit:
@alexmisk 
I've checked the code and CI logs, and LGTM! 

That's great job! Thanks a lot! 🎉 😄

PR #74, opened 2021-01-04
Fix pytest error

adsharma:
Without this fix, a variant of this issue was seen
https://github.com/pytest-dev/pytest/issues/638


yukinarit:
@adsharma Thank you!!

PR #87, opened 2021-02-10
Improve Union type support

### LINKED ISSUE #60: Write more tests around Union

ydylla:
This tries to add support for Unions with complex types like `UUID` or `IPv4Address`.

The basic idea is to use the code generation to add function for each Union the dataclass uses.
This union function then tries to serialize or deserialize the field according to the order of types defined in the Union.
The first one that works is used.

Using the code generation for this has the advantage that the user can influence the order of checks and that only types are checked that are defined in the Union.
Which should be faster, then always checking against all types that pyserde supports.

Currently, not all tests work. I have not found a good solution for nested Unions or container types within Unions, e.g. `Union[List[int],List[str]]`.
We probably need some form of recursion.

Some notable other changes:
* Implemented `is_set` in compat.py
* The rendered dict & iter functions now define a variable `serde_scope` which I needed to pass it to the union functions.
* The render functions now render `NoneType` as `None`. I think this broke the `test_union_optional` test, but I still have to look into it. 

@yukinarit If you have time, I would appreciate a review and your opinion about the general idea?
Also I noticed you increased the version to 0.2.3 so only at patch level.
Do you use [semantic versioning](https://semver.org)? Then it should be 0.3.0 in my opinion because the new type support is a feature :smile:.

Closes #60


serde/de.py
```diff
@@ -569,6 +594,44 @@ def {{func}}(data, reuse_instances = {{reuse_instances_default}}):
...
+    error = "input is not of type {{t.__name__}}"
+  {% else %}
+  try:
+    return {{t|arg|rvalue}}
+  except Exception as e:
+    error = str(e)
+  {% endif %}
+  {% endfor %}
+  raise SerdeError("Can not deserialize " + repr(obj) + " of type " + typename(type(obj)) + " into {{union_name}}. Reason: " + error)
+    """
```
yukinarit, line 622: General idea of the implementation looks good to me! 😄  

I had tried to implement Union support back in [pyserde 0.0.7](https://github.com/yukinarit/pyserde/commit/8c4503c31d37bf074381db723ff463ba13a0b88f), but my implementation wasn't generation based. It resolved a matching type in Union at runtime, thus it wasn't efficient enough.

yukinarit (commented):
@ydylla 
Thank you for the contribution! Sorry for the delay in getting back to you because it was a holiday in Japan 😉  

I left review comments. Please check 👍

yukinarit:
> Do you use [semantic versioning](https://semver.org)? Then it should be 0.3.0 in my opinion because the new type support is a feature 😄.
> 
> Closes #60

Yes I treat pyserde versioning as `0.MAJOR.MINOR` as discussed in [this issue](https://github.com/semver/semver/issues/363#issue-220038873), but am not 100% sure this is right. How do you treat versioning before 1.0.0? 🤔

serde/core.py
```diff
@@ -217,3 +221,26 @@ def conv(f: Field, case: Optional[str] = None) -> str:
...
+    """
+    name = ""
+    for arg in union_args:
+        # handles container types like List,Tuple & Dict
+        if isinstance(arg, _GenericAlias):
+            name += f"{arg._name}_{union_func_suffix(type_args(arg))}"
+        else:
+            name += arg.__name__
+        name += "_"
+    return name + "_"
```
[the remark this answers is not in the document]
ydylla: Sorry I always forget this 😄 
Maybe I should enable auto new line again.

ydylla:
> Yes I treat pyserde versioning as 0.MAJOR.MINOR as discussed in this issue, but am not 100% sure this is right. How do you treat versioning before 1.0.0? 🤔

In my projects I do not treat 0.x.x specially. I [start with 0.1.0](https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase) and then increase minor 0.2.0 for features and patch 0.2.1 for fixes.
So once I need to assign a version and have other people using it I follow semantic versioning.
If I have to break something I would increase the major version despite [semver allowing breaking changes in 0.x.x](https://semver.org/#spec-item-4)

yukinarit:
> If I have to break something I would increase the major version despite semver allowing breaking changes in 0.x.x

Oh! I didn't know sember allows breaking change before 1.0.0 😲 
I still don't want to update the major version because pyserde is not stable enough.

Anyway, I will change it from 0.2.3 to 0.3.0 maybe after a few more MRs are merged. 👍  
Let me know If you need a new version pyserde in PyPI

ydylla:
Hi @yukinarit,
this is now ready to be reviewed, so if you have time please take a look.

I am not sure if you like 431c9146e31210a4d3e999dd2456667a1ea3acb0 but I thought it would be nice (and cleaner) to use the serde scope like an actual scope, which contains all serde internals. This way we do not need several `__foo__` names to hide the internal functions. In the generated functions the scope is now always available as `serde_scope` (for jinja and within the functions).

If you do not like it, it is no problem to get the union support working without it. So do not feel obliged to merge this.

serde/compat.py
```diff
@@ -194,7 +221,7 @@ def is_bare_list(typ) -> bool:
     >>> is_bare_list(List)
     True
     """
-    return is_list(typ) and typ in (List, list)
+    return typ in (List, list)
```
yukinarit, line 224: What's the reason for removing `is_list(typ)` 🤔

yukinarit (commented):
Hi @ydylla  
I don't mind replacing scope object with dataclass. I left a few comments in order to understand the intent of the code. Please answer to the question when you have time.

Thanks!

serde/compat.py
ydylla: It does not do anything.   
The `typ in (List, list)` check is stricter and determines the output of that statement completely.

yukinarit:
LGTM

ydylla:
🥳 Thanks for the merge.

yukinarit:
No problem! Thanks for the contribution! 🙂

PR #102, opened 2021-03-10
Fix case conversion for optional

### LINKED ISSUE #100: pyserde doesn't convert case for optional type when "rename_all" attributes specified
I expect to get  
`From Json: Foo(name='Pikachu', no=25)` from the below code, but it actually produces 
`From Json: Foo(name='Pikachu', no=None)`.

```python
@deserialize(rename_all='pascalcase')
@serialize(rename_all='pascalcase')
@dataclass()
class Foo:
    name: str
    no: Optional[int] = None


s = '{"Name": "Pikachu", "No": 25}'
```

---

The generated code:

```python
$ python -m serde.inspect examples/rename_all.py Foo

...
def from_dict(data, reuse_instances = True):
  if reuse_instances is Ellipsis:
    reuse_instances = True
  Foo = serde_scope.types['Foo']
  if data is None:
    return None
  return cls(
  data["Name"],
  (data["no"]) if data.get("no") is not None else serde_scope.defaults["no"],
  )
```

Apparently pyserde doesn't convert the `"no"` to pascalcase `"No"`  for this case.

yukinarit:
Close #100 

* Add a new property `conv_name`
* Use `conv_name` whereever the name after conversion is needed

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -234,6 +234,14 @@ class Field:
         """
         return f'{field.name}_{name}'
 
+    @property
+    def conv_name(self) -> str:
+        """
+        Get an actual field name which `rename` and `rename_all` conversions
+        are made. Use `name` property to get a field name before conversion.
+        """
+        return conv(self, self.case)
+
 
 def fields(FieldCls: Type, cls: Type) -> Iterator[Field]:
     return iter(FieldCls.from_dataclass(f) for f in dataclasses.fields(cls))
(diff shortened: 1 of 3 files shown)


yukinarit:
@ydylla  Please review 🙂

serde/core.py
```diff
@@ -250,7 +258,7 @@ def conv(f: Field, case: Optional[str] = None) -> str:
             raise SerdeError(
                 (f"Unkown case type: {f.case}." f"Pass the name of case supported by 'stringcase' package.")
```
ydylla, line 259: I think the auto formatting was a bit confused here, just
```
raise SerdeError(
    f"Unkown case type: {f.case}. Pass the name of case supported by 'stringcase' package."
)
```
should be fine.

ydylla: And on line [256](https://github.com/yukinarit/pyserde/pull/102/files#diff-217837e513e75f90cbaf8744df290da74e5bfc091c4deefd4b8d28705920f37fR256) just `casef = getattr(stringcase, case, None)` should be fine too. The if statement above already covers the or case.

ydylla:
@yukinarit Looks good to me. I only noticed some formatting hiccups.

serde/core.py
yukinarit: OK, fixed!

PR #103, opened 2021-03-10
CI: pin runner OS versions

alexmisk:
Well, it's almost always good to pin your dependencies, even CI runner OS versions :-) 

Recent migration of GitHub runners from Ubuntu 18.04 to 20.04 resulted in the situation in which runner with the newer Ubuntu version was restoring cache from the older one because they have the same cache key ("ubuntu-latest-..."). 

I pinned all OS versions in the pipeline to prevent this in future.


yukinarit:
LGTM

PR #106, opened 2021-03-21
Fix dates in CHANGELOG.md

alexmisk:


yukinarit:
0.3.0 was actually released on March 20, 2021? Could you also fix?

alexmisk:
> 0.3.0 was actually released on March 20, 2021? Could you also fix?

Done!

yukinarit:
Thanks!

PR #118, opened 2021-04-23
feat: Improve error description for union type

yukinarit:
It's convenient if you can see what error each type in union received during deserialization.

Given this class,

```python
@deserialize
@serialize
@dataclass
class A:
    v: Union[IPv4Address, UUID]
```

Previously, this kind of error description are returned.

```
Can not deserialize 'not-ip-or-uuid' of type str into Union[IPv4Address, UUID]. Reason: badly formed hexadecimal UUID string
```

However, it's impossible to know why the type IPv4Address failed.

With this change, you can receive the following description instead.

```
Can not deserialize 'not-ip-or-uuid' of type str into Union[IPv4Address, UUID]
Reason:
Exhausted all types
 Failed to deserialize into IPv4Address: Expected 4 octets in 'not-ip-or-uuid
 Failed to deserialize into UUID: badly formed hexadecimal UUID string
```

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -646,21 +646,21 @@ def {{func}}(data, reuse_instances):
   # create fake dict so we can reuse the normal render function
   fake_dict = {"fake_key":data}
 
-  error = "Exhausted all types"
+  errors = []
   {% for t in union_args %}
   {% if t | is_primitive or t | is_none %}
   if isinstance(data, {{t.__name__}}):
     return {{t|arg|rvalue}}
   else:
-    error = "input is not of type {{t.__name__}}"
+    errors.append("input is not of type {{t.__name__}}")
   {% else %}
   try:
     return {{t|arg|rvalue}}
   except Exception as e:
-    error = str(e)
+    errors.append(f' Failed to deserialize into {{t.__name__}}: {e}')
   {% endif %}
   {% endfor %}
-  raise SerdeError("Can not deserialize " + repr(data) + " of type " + typename(type(data)) + " into {{union_name}}. Reason: " + error)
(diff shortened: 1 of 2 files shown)


serde/de.py
```diff
@@ -646,21 +646,21 @@ def {{func}}(data, reuse_instances):
   # create fake dict so we can reuse the normal render function
   fake_dict = {"fake_key":data}
 
-  error = "Exhausted all types"
+  errors = ["Exhausted all types"]
```
ydylla, line 649: I think without the "Exhausted all types" message and an extra space in front of the primitives error message it would look even nicer.
```
Can not deserialize 'not-ip-or-uuid' of type str into Union[int, bool].
Reasons:
 input is not of type int
 input is not of type bool
```

serde/de.py
```diff
@@ -646,21 +646,21 @@ def {{func}}(data, reuse_instances):
...
   {% else %}
   try:
     return {{t|arg|rvalue}}
   except Exception as e:
-    error = str(e)
+    errors.append(f' Failed to deserialize into {{t.__name__}}: {e}')
   {% endif %}
   {% endfor %}
-  raise SerdeError("Can not deserialize " + repr(data) + " of type " + typename(type(data)) + " into {{union_name}}. Reason: " + error)
+  raise SerdeError("Can not deserialize " + repr(data) + " of type " + typename(type(data)) + " into {{union_name}}.\\nReason:\\n" + "\\n".join(errors))
```
ydylla, line 663: We should call it "Reason**s**" since we always have multiple errors.
Except when one uses a single argument union, which should not happen 😄

ydylla (approved):
Great idea 👍 
I have only added some comments to further improve the formatting.

yukinarit:
@ydylla All addressed! Please merge if it's ok 🙂

PR #123, opened 2021-05-01
Some bugfixes

### LINKED ISSUE #113: Unable to deserialize nested types of union and tuple
**Python 3.9 pyserde[toml]==0.3.1**

```python
from typing import Union
from serde import toml, deserialize


@deserialize
@dataclass
class TestClass:
    a: Union[None, Union[int, float], tuple[Union[int, float], Union[int, float]]]


if __name__ == '__main__':
    test_class = toml.from_toml(TestClass, 'a = [4, 1]')
    print(test_class)

```

Running the above code will throw an error:

```serde.core.SerdeError: Can not deserialize [4, 1] of type List into Union[NoneType, int, float, Tuple[Union[int, float], Union[int, float]]]. Reason: 'union_de_int_float'```

ydylla:
Hi @yukinarit,
I fixed some bugs. Mainly regarding #113

1. Unions with other unions in container types should work now.
2. I also fixed the tuple bug. The problem was that single element tuples where rendered as `("a")`. Which becomes just `"a"` instead of a tuple. The fix was always adding a trailing comma like so `("a",)`.
3. The last bug was that the `convert_sets` argument was not correctly passed to the rendered union functions. Which prevented the use of other dataclasses in a container type within a union.

Closes #113

### DIFF
--- a/serde/compat.py
+++ b/serde/compat.py
@@ -159,6 +159,8 @@ def iter_unions(cls: Type) -> Iterator[Type]:
     """
     if is_union(cls):
         yield cls
+        for arg in type_args(cls):
+            yield from iter_unions(arg)
     if is_dataclass(cls):
         for f in fields(cls):
             yield from iter_unions(f.type)
--- a/serde/de.py
+++ b/serde/de.py
@@ -503,11 +503,11 @@ class Renderer:
         ... @dataclass
         ... class Foo: pass
         >>> Renderer('foo').render(DeField(Tuple[str, int, List[int], Foo], 'd', datavar='data'))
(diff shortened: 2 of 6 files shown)


yukinarit:
@ydylla  
Thanks for the fix! LGTM. 

I got the same pip error and am looking into it: https://stackoverflow.com/a/67273650

yukinarit:
@ydylla  I've pushed a temporary fix for pip error. Could you rebase to master?

ydylla:
> 
> 
> @ydylla I've pushed a temporary fix for pip error. Could you rebase to master?

Sure, done 😄

yukinarit:
Thanks! 🙂

PR #126, opened 2021-05-08
Add support for lazy type annotations PEP563 (#112)

### LINKED ISSUE #112: Support PEP563 lazy type evaluation
Adding `from __future__ import annotations` to examples/simple.py raises an exception.

```python
from __future__ import annotations

from serde import deserialize, serialize


@deserialize
@serialize
@dataclass
class Foo:
    i: int
    s: str
    f: float
    b: bool


def main():
    f = Foo(i=10, s='foo', f=100.0, b=True)
    print(f"Into Json: {to_json(f)}")

    s = '{"i": 10, "s": "foo", "f": 100.0, "b": true}'


if __name__ == '__main__':
    main()
```

```
Into Json: {"i": null, "s": null, "f": null, "b": null}
  File "/Users/yukinari/repos/pyserde/examples/simple.py", line 27, in <module>
    main()
  File "/Users/yukinari/repos/pyserde/examples/simple.py", line 23, in main
  File "/Users/yukinari/repos/pyserde/serde/json.py", line 29, in from_json
  File "/Users/yukinari/repos/pyserde/serde/de.py", line 274, in from_dict
  File "/Users/yukinari/repos/pyserde/serde/de.py", line 211, in from_obj
    return serde_scope.funcs[FROM_DICT](o, reuse_instances=reuse_instances)
  File "<string>", line 16, in from_dict
  File "/Users/yukinari/repos/pyserde/serde/core.py", line 82, in raise_unsupported_type
    raise SerdeError(f"Unsupported type: {typename(type(obj))}")
serde.core.SerdeError: Unsupported type: int
```

ydylla:
Hi @yukinarit,
I added basic support for `from __future__ import annotations`.

Do you have any ideas how we could tell pytest to skip test_lazy_type_evaluation.py when using python 3.6 since the annotations future does not exist there?

Closes #112

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,8 @@
+# see https://docs.pytest.org/en/latest/example/pythoncollection.html#customizing-test-collection
+import sys
+
+collect_ignore = ["setup.py"]
+
+if sys.version_info[:2] == (3, 6):
+    # skip these tests for python 3.6 because it does not support PEP 563
+    collect_ignore.append("tests/test_lazy_type_evaluation.py")
--- a/serde/compat.py
+++ b/serde/compat.py
@@ -3,8 +3,9 @@ Module for compatibility.
 """
 import dataclasses
 import enum
+import sys
 import typing
-from dataclasses import fields, is_dataclass
+from dataclasses import is_dataclass
 from itertools import zip_longest
 from typing import Any, Dict, Iterator, List, Optional, Set, Tuple, Type, TypeVar, Union
 
@@ -13,6 +14,13 @@ import typing_inspect
 __all__: List = []
 
 T = TypeVar('T')
+
+
+# moved SerdeError from core.py to compat.py to prevent circular dependency issues
+class SerdeError(TypeError):
+    """
+    Serde error class.
+    """
 
 
 def get_origin(typ):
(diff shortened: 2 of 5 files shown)


ydylla:
Yes I am aware of `pytest.mark.skipif`, sorry I should have mention that.
The Problem is that the python 3.6 tests fail at `from __future__ import annotations` which has to be the first line in a file.
I think we need a way to ignore the complete `test_lazy_type_evaluation.py` file in the pipeline. But I have not much experience with github wokflows. An ugly hack could be to just delete the file before starting pytest 😄 

I also forgot to mention that I added tests for forward references with this.
We can not really support forward references, but maybe we should mention the workaround in the docs/readme. Which basically is call `deserialize` & `serialize` only after all classes that are used in the forward reference are visible globally.
```python
@dataclass
class A:
  v: 'B'

@serialize
@deserialize
@dataclass
class B:
  v: int

# B is now visible so we can setup A
serialize(A)
deserialize(A)
```

yukinarit:
@ydylla 
Oh! Now I understand the problem! 

I found a workaround in pydantic project. Could you check it is applicable to our case?

```python
kip_pre_37 = pytest.mark.skipif(sys.version_info < (3, 7), reason='testing >= 3.7 behaviour only')


@skip_pre_37
def test_postponed_annotations(create_module):
    module = create_module(
        # language=Python
        """
from __future__ import annotations
from pydantic import BaseModel
class Model(BaseModel):
    a: int
"""
    )
    m = module.Model(a='123')
    assert m.dict() == {'a': 123}
```

https://github.com/samuelcolvin/pydantic/blob/master/tests/test_forward_ref.py#L8-L24

yukinarit (commented):
LGTM!

ydylla:
Hi @yukinarit,
I tried to dynamically create modules like pydantic, but I could not get it to work properly.
Also I did not like that a lot of code only existed as string, which makes debugging very annoying.
Instead, I pushed a fix that just deletes the whole file for python 3.6.
Which was way easier and should probably be fine until 2021-12-23 when python 3.6 goes out of support anyways.

If you have no objections you can merge this.

ydylla:
Thanks @yukinarit,
this is definitely better (the intended way to do this).

yukinarit (approved):
LGTM!

yukinarit:
@ydylla  
Thanks! Let me know if you need the latest pyserde on PyPI now.

ydylla:
Currently, not necessarily. I already use it via git installation.

PR #134, opened 2021-05-29
style: Fix some mypy/flake8 warnings

yukinarit:


serde/se.py
```diff
@@ -590,7 +591,7 @@ def primitive(self, arg: SeField) -> str:
         """
         return f'{arg.varname}'
 
-    def str(self, arg: SeField) -> str:
+    def string(self, arg: SeField) -> str:
```
yukinarit, line 594: @ydylla  Renamed `str` to `string` because `str` confuses mypy

PR #135, opened 2021-06-04
feat: Implement custom class (de)serializer

yukinarit:
Hi @ydylla  

I've implemented class-level custom (de)serializer. PTAL

I would like to hear your opinion about the attribute names `@deserialize(deserialize=XXX)` and `@serialize(serialize=YYY)`. `serialize` part is repeating in `@serialize` and `serialize=YYY`. They don't look awkward? 🤔  I picked those names in accordance with `serde_serialize`/`serde_deserialize` field attributes. If you know any better name to suggest, I'd appreciate it. Thanks!

```python
def serializer(cls, o):
    if cls is datetime:
        return o.strftime('%d/%m/%y')
    else:
        raise SerdeSkip()

def deserializer(cls, o):
    if cls is datetime:
        return datetime.strptime(o, '%d/%m/%y')
    else:
        raise SerdeSkip()

@deserialize(deserialize=deserializer)
@serialize(serialize=serializer)
@dataclass
class Foo:
    i: int
    dt1: datetime
    dt2: datetime
```

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -85,7 +85,7 @@ def serde_custom_class_deserializer(cls: Type, datavar, value, custom: Deseriali
         return default()
 
 
-def default_deserialize(_cls: Type, obj):
+def default_deserializer(_cls: Type, obj):
     """
     Marker function to tell serde to use the default deserializer. It's used when custom deserializer is specified
     at the class but you want to override a field with the default deserializer.
@@ -96,7 +96,7 @@ def deserialize(
     _cls: Type[T] = None,
     rename_all: Optional[str] = None,
     reuse_instances_default: bool = True,
-    deserialize: Optional[DeserializeFunc] = None,
+    deserializer: Optional[DeserializeFunc] = None,
 ) -> Type[T]:
     """
     `deserialize` decorator. A dataclass with this decorator can be deserialized
@@ -149,7 +149,7 @@ def deserialize(
         g['typename'] = typename  # used in union functions
         if deserialize:
             g['serde_custom_class_deserializer'] = functools.partial(
-                serde_custom_class_deserializer, custom=deserialize
+                serde_custom_class_deserializer, custom=deserializer
             )
 
         # Collect types used in the generated code.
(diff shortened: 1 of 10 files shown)


serde/de.py
```diff
@@ -367,12 +399,13 @@ class Renderer:
     """
 
     func: str
+    custom: Optional[DeserializeFunc] = None  # Custom class level deserializer.
 
     def render(self, arg: DeField) -> str:
         """
         Render rvalue
         """
-        if arg.serialize:
```
yukinarit, line 375: This is a bug in the previous MR.

serde/de.py
```diff
@@ -437,7 +470,14 @@ def render(self, arg: DeField) -> str:
             elif has_default_factory(arg):
                 res = f'({res}) if {exists} else serde_scope.defaults["{arg.name}"]()'
 
-        return res
+        if self.custom and not arg.deserialize:
+            # The function takes a closure in order to execute the default value lazily.
+            return (
+                f'serde_custom_class_deserializer({arg.type.__name__}, {arg.datavar}, {arg.data}, '
+                f'default=lambda cls, data: {res})'
```
ydylla, line 477: Why do you pass arguments to the lambda here?
I think a lambda without arguments would work fine and is cleaner.

serde/se.py
```diff
@@ -507,15 +526,21 @@ def render(self, arg: SeField) -> str:
...
+            res = "None"
         elif arg.type is Any:
-            return f"to_obj({arg.varname}, True, False, False)"
+            res = f"to_obj({arg.varname}, True, False, False)"
+        else:
+            res = f"raise_unsupported_type({arg.varname})"
+
+        # Custom field serializer overrides custom class serializer.
+        if self.custom and not arg.serialize:
+            return f'serde_custom_class_serializer({arg.type.__name__}, {arg.varname}, default={res})'
```
ydylla, line 541: For serialize the default value should probably also be a lambda.

ydylla (commented):
Hi @yukinarit,
I have reviewed your code.
But I don't really have an idea for a better name either. Maybe `custom` or `customize`?

serde/de.py
yukinarit: You're right. It works 😄

serde/se.py
yukinarit: True. I will take a look.

yukinarit:
@ydylla  

For the naming, here is the list of candidates

| |  Class   | Field   | Comment  |
|-|---|---|---|
|1| `serialize`/`deserialize`  | `serde_serialize`/`serde_deserialize`  | This is the current implementation.  |
|2| `serializer`/`deserializer`  | `serde_serializer`/`serde_deserializer` | Slightly better than `1` because the name they are different from decorator names? 🤔  |
|3|  `custom`/`custom` | `serde_custom`/`serde_custom`  | Also looks good to me.  |
|4|  `customize`/`customize` | `serde_customize`/`serde_customize`  |  I rather prefer `3` because its' shorter 😄  |
|5|  `converter`/`converter` | `serde_converter`/`serde_converter`  |   |

NOTE:
* I want the same naming for the class/field attributes.

yukinarit:
@ydylla  I ponder the names, and think I am going to pick No.2 `serializer/deserializer`, `serde_serializer/serde_deserializer` 👍

yukinarit:
@ydylla  Addressed all the comments! Please merge if you think it's ok! Thanks!

ydylla (approved):
LGTM

yukinarit:
Thank you!

PR #144, opened 2021-07-29
Migrate to poetry

alexmisk:
Here is an almost working draft, I'll leave a few comments later today. You can examine a build generated by Poetry here: https://test.pypi.org/project/pyserde-test/


alexmisk:
A few notes on the changes. I've somewhat mitigate problems with Poetry so one can manage virtualenvs, dependencies, build process and publishing with it now. 

Tried to fully mimic current pipenv / setup.py behavior, but faced these issues:
- Poetry config become bloated: to support separation of dependencies both after cloning project from GitHub and installing it from PyPi we need to double-declare deps in several config sections (see https://github.com/python-poetry/poetry/issues/1416). Config can be trimmed though if we drop support for installing test deps via PyPi (`pip install pyserde[test]`)
- Pre-commit refused to pick up config for black and isort from pyproject.toml after migration, so I moved it to .pre-commit-config.yaml (but it’s for the better, actually)
- Pre-commit minimal Python version is 3.6.1, so I had to bump it in the pyproject.toml. Poetry refuses to install deps with 3.6
- Poetry currently has an issue with Python 3.10 (see https://github.com/python-poetry/poetry/issues/4210) so I removed 3.10-dev from CI pipeline. It’s certainly a temporary thing
- I've to remain Pipfile in examples folder untouched because tomlfile.py depends on it

Everything else seems fine. Waiting for your comments! :-)

yukinarit:
Hi @alexmisk 

Sorry for the delay! 

> Config can be trimmed though if we drop support for installing test deps via PyPi

I don't mind dropping `pip install pyserde[test]`. I don't use it, and probably nobody uses it 😄

alexmisk:
Hi!

Extra deps removed, OS name changed :-) PR is ready to merge. 

Yet another detail: as you may know, Poetry tracks package version in pyproject.toml and has a convenient way to change it. For example:

``` bash
❯ poetry version minor
Bumping version from 0.4.0 to 0.5.0
❯ poetry version patch
Bumping version from 0.5.0 to 0.5.1
```
More on that: https://python-poetry.org/docs/cli/#version

So maybe there is no need to keep version number in `serde/__init__.py` anymore. But it's up to you, of course.

yukinarit:
LGTM! 🎉

yukinarit:
I don't mind deleting version number in `serde/__init__.py`.

PR #159, opened 2021-10-18
refactor: Use backports-datetime-fromisoformat for python 3.6

yukinarit:
It's better to use [backports-datetime-fromisoformat](https://pypi.org/project/backports-datetime-fromisoformat/) than implementing own? This will drop the support of pypy 3.6 though

CC @ydylla


yukinarit:
@ydylla 
Thanks for reviewing!

PR #169, opened 2021-11-10
fix: Add types in typing module to scope

### LINKED ISSUE #167: Missing types with custom class deserializers
The code generation doesn't put `Optional` or `Union` in scope when calling a custom class deserializer. This results in a `NameError` during deserialization.

Example
```py
from typing import Optional

from serde import deserialize, from_dict


def deserialize_maybe_str(_typ, value):
    return value


@deserialize(deserializer=deserialize_maybe_str)
@dataclass
class CustomDe:
    maybe_str: Optional[str]


if __name__ == "__main__":
    data = {"maybe_str": "bye"}
    print(f"from_dict({data}) = {from_dict(CustomDe, data)}")
```
Output
```
  File "C:\Users\mauve\dev\modlunky2\src\union_demo.py", line 19, in <module>
    print(f"from_dict({data}) = {from_dict(CustomDe, data)}")
  File "C:\Users\mauve\dev\modlunky2\.venv\lib\site-packages\serde\de.py", line 319, in from_dict
  File "C:\Users\mauve\dev\modlunky2\.venv\lib\site-packages\serde\de.py", line 253, in from_obj
    return serde_scope.funcs[FROM_DICT](o, reuse_instances=reuse_instances)
  File "<string>", line 10, in from_dict
NameError: name 'Optional' is not defined
```

yukinarit:
Closes #167

### DIFF
--- a/serde/se.py
+++ b/serde/se.py
@@ -18,6 +18,7 @@ from .compat import (
     SerdeSkip,
     is_bare_dict,
     is_bare_list,
+    is_bare_opt,
     is_bare_set,
     is_bare_tuple,
     is_dict,
@@ -46,6 +47,7 @@ from .core import (
     add_func,
     conv,
     fields,
+    filter_scope,
     is_instance,
     logger,
     raise_unsupported_type,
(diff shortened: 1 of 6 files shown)


yukinarit:
This MR fixes the #167 but I would like to find a more elegant way 🤔

serde/se.py
```diff
@@ -174,6 +174,7 @@ def wrap(cls: Type):
         g['typename'] = typename  # used in union functions
         g['is_instance'] = is_instance  # used in union functions
         g['to_obj'] = to_obj
+        g['Optional'] = Optional
```
yukinarit, line 177: I feel like we should import `typing` module into the scope instead of `Optional`, `Union` ..etc

PR #172, opened 2021-11-18
refactor: Remove type references from SerdeScope

yukinarit:
References to types used in generated functions are now added to
global scope of `exec` function. No need to have `types` property
in SerdeScope object anymore.


yukinarit:
Hi @ydylla 

Is there any issue if I remove `types` from `SerdeScope`? PTAL

ydylla:
> Is there any issue if I remove types from SerdeScope?

No I think everything should work fine, and it is less redundant 👍

yukinarit:
@ydylla Thanks!

PR #197, opened 2022-01-28
core: use black formatting only if debug is enabled

tardyp:
There is currently  a small issue on black which makes it try to rebuild is cache forever.

```
Generating grammar tables from site-packages/blib2to3/Grammar.txt
Writing grammar tables to  Grammar3.10.1.final.0.pickle
Writing failed: [Errno 2] No such file or directory: 'XXX/black/21.10b0/tmpi9dlr3sd'
```
This is how I saw that the generated code was actually using black formatting before compile.
Not sure why it is necesarry, but probably it is only needed for debugging.

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -162,12 +162,14 @@ def gen(code: str, globals: Dict = None, locals: Dict = None) -> str:
     """
     A wrapper of builtin `exec` function.
     """
-    try:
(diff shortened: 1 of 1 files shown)


yukinarit (approved):
@tardyp Thanks!

PR #198, opened 2022-01-30
Strict type checking

### LINKED ISSUE #164: No error if value has the wrong type
When converting a toml file which has the correct fields, but the wrong type, the code will just succeed:
```python

from serde import deserialize, serialize
from serde.toml import from_toml

@deserialize
@serialize
@dataclass
class Test:
    foo: float

test = from_toml(Test, """
        foo = "test"
        """
        )
print(test)
```

This will just print `Test(foo='test')` with `foo` as a `str`. I'd expect the code to raise an error indicating that the field has the wrong type.

yukinarit:
`pyserde` provides 3 kinds of type checks.

### `NoCheck`

This is the default behavior until pyserde v0.8.3. No type coercion or checks are applied. If a user puts a wrong value, pyserde doesn't complain anything.

```python
@serde
@dataclass
class Foo
    s: str

foo = Foo(10)
# pyserde doesn't complain anything. {"s": 10} will be printed.
print(to_json(foo, type_check=NoCheck)) 

```

### `Coerce`

Type coercion is not yet implemented but this will be the default behavior in pyserde v0.9.0 onward.

```python
foo = Foo(10)
# pyserde automatically coerce the int value 10 into "10".
# {"s": "10"} will be printed.
print(to_json(foo, type_check=Coerce)) 
```

Not yet implemented.

### `Strict`

Strict type checking is to check every value against the declared type.

```python
foo = Foo(10)
# pyserde checks the value 10 is instance of`str`.
# SerdeError will be raised in this case because of the type mismatch.
print(to_json(foo, type_check=Strict)) 
```

Closes #164


gitpushdashf:
I'm very interested in this! Would be quite handy. Might also slow things down, but in some cases this might be able to be a pydantic alternative.

yukinarit:
@gitpushdashf Yes, I am trying to push this feature soon 🚀

serde/de.py
```diff
@@ -6,6 +6,7 @@
 import abc
 import dataclasses
 import functools
+import traceback
 import typing
```
gitpushdashf, line 10: For Python 3.7 support, Literal has to come from the typing-extensions package.

Could maybe do something like:

```
try:
    from typing import Literal
except ImportError:
    from typing_extensions import Literal
```

Or

https://github.com/python/mypy/issues/8155

```
import sys

if sys.version_info[:2] >= (3, 8):
    from typing import Literal
else:
    from typing_extensions import Literal
```

yukinarit:
@gitpushdashf Done. Would you please help me review this?

PR #203, opened 2022-02-27
Support generics

yukinarit:
Hi @jennydaman!

It's tricky but possible to support Generics. Here is POC implementation.

Since I am not a heavy user of `typing.Generics`, I would like to ask you to confirm few things. It would be great if you could know.

1. What is the best way to check the type inheriting `typing.Generics`.

I use this code in the POC.
```python
def is_generic(typ) -> bool:
    return typing.get_origin(typ) is not None
```

2. When a generic type is not specified like `MyClass` instead of `MyClass[Nested]`, what do you expect the type of `thing`? Is it any? 🤔

### DIFF
--- a/tests/test_union.py
+++ b/tests/test_union.py
@@ -1,14 +1,14 @@
 import logging
 from dataclasses import dataclass
 from ipaddress import IPv4Address
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, Generic, List, Optional, Tuple, TypeVar, Union
 from uuid import UUID
 
 import pytest
 
-from serde import SerdeError, from_dict
+from serde import SerdeError, from_dict, from_tuple
 from serde import init as serde_init
-from serde import logger, serde, to_dict
+from serde import logger, serde, to_dict, to_tuple
 from serde.json import from_json, to_json
 
 logging.basicConfig(level=logging.WARNING)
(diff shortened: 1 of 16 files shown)


jennydaman:
@yukinarit I really appreciate you for considering this! I'm not home right now so sending what I know from my phone.

So generics are really tricky, I'm going to be real -- if I were you, I'd mark my request as "wontfix" but keep it as an interesting thing in the back of my mind.

1. I don't know. I developed a recursive solution to get a _bound_ generic type of a class, or any of its subclasses. Not sure if there is a better way to get the generic other than traversing the subclasses.

https://github.com/FNNDSC/chrisomatic/blob/28e00c999a020c9199c6d8a5f26a22c9449b529a/chris/common/client.py#L287-305

2. `typing.Any` for unbound generics, or the bound type (some superclass). (In the latter case, PyCharm still tells me `Any`, suggesting that inspection of a bound generic is not trivial.)

yukinarit:
Thanks for you input, @jennydaman 

> So generics are really tricky, I'm going to be real -- if I were you, I'd mark my request as "wontfix" but keep it as an interesting thing in the back of my mind.

NP, I also wanted to support Generics someday. Now It's time  😄

yukinarit:
Hi @jennydaman 

I think you can use `typing.Generic` now! Please play around with `support-generics` branch?

tests/test_union.py
```diff
@@ -327,6 +327,44 @@ class A:
...
+    class B(Generic[T, U]):
+        v: Union[Foo[T], Bar[U]]
+
+    b = B[int, str](Foo[int](10))
+    assert b == from_dict(B[int, str], to_dict(b))
+    assert b == from_tuple(B[int, str], to_tuple(b))
+
+    b = B[Foo[int], Bar[str]](Foo(Foo(10)))
+    assert {'v': {'v': {'v': 10}}} == to_dict(b)
+    # TODO Nested parameterized generics still doesn't work
```
jennydaman, line 364: Stale TODO? This test is passing

jennydaman:
@yukinarit it works really well!

tests/test_union.py
yukinarit: This assert is supposed to be `==` not `!=`.

I will fix it to `==` and comment out.
```python
# assert b == from_dict(B[Foo[int], Bar[str]], to_dict(b))
```

yukinarit:
@jennydaman Thanks!

PR #209, opened 2022-03-24
feat: Support numpy types

### LINKED ISSUE #145: Add example for serializing Numpy data
After playing for some time, it would be nice to have an example for serialization with a Numpy type. I have given a starting example below. The essence is:
- Specify type annotation with `numpy.typing`
- Use `__post_init__` to convert a `List[float]` to the annotated type
- When (de)serializing, convert to a serializable object type

**examply.py**
```python
from dataclasses import dataclass, field
from serde import serialize, deserialize
import numpy as np
import numpy.typing as npt


@serialize
@deserialize
@dataclass
class Foo:
    a: npt.NDArray[np.float64] = field(
        metadata={
            'serde_serializer': lambda x: x.tolist(),
            'serde_deserializer': lambda x: np.array(x, dtype=np.float64),
        }
    )

    def __post_init__(self):
        self.a = np.array(self.a, dtype=np.float64)

# Create using a list
f = Foo(a=[1., 0.])

# Check type of "f.a"
print(type(f.a))

# Serialize
print(to_json(f))

# Deserialize
s = '{"a": [1.0, 2.0]}'
print(from_json(Foo, s))
```

**python example.py**
```bash
<class 'numpy.ndarray'>
{"a": [1.0, 0.0]}
Foo(a=array([1., 2.]))
```

**Pit falls**
Ideally, `numpy.ndarray` could serialize itself. If one removes the `serde_*` metadata, the following error is found:
```bash
serde.compat.SerdeError: Unsupported type: ndarray
```
It is very nice to include the numpy type in the dataclass instead of `List[float]` because then all the methods can be called directly. (e.g. `f.a.max()`). Maybe there is a better solution to avoid all the extra field methods?

Any feedback would be appreciated and whether this is a worthwhile contribution.

kmsquire:
* Adds support for numpy types specified in dataclasses, converting
  them to and from standard python types as needed.
* ~Note that this commit does NOT handle the case where a numpy
  type is misrepresented as a standard python type.  In most cases,
  these values will fail to serialize, or may be serialized
  incorrectly.~

The second commit now supports serialization of untyped or incorrectly typed numpy values (i.e., where the type of a variable is a regular Python type, but the contents are a numpy scalar or array) *for json and toml only*.  Deserialization is unaffected in these cases--i.e., the value will be deserialized to the declared (non-numpy) type.  (Deserialization to declared numpy types is handled in the first commit).

It should be possible to add support for yaml and toml, but the mechanism for doing so is different and not as obvious.

Closes #145

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,11 +32,13 @@ jinja2 = "*"
 msgpack = { version = "*", markers = "extra == 'msgpack' or extra == 'all'", optional = true }
 toml = { version = "*", markers = "extra == 'toml' or extra == 'all'", optional = true }
 pyyaml = { version = "*", markers = "extra == 'yaml' or extra == 'all'", optional = true }
+numpy = { version = "*", markers = "extra == 'numpy' or extra == 'all'", optional = true }
 
 [tool.poetry.dev-dependencies]
 pyyaml = "*"
 toml = "*"
 msgpack = "*"
+numpy = "*"
 flake8 = "*"
 pytest = "*"
 pytest-cov = "*"
@@ -51,10 +53,12 @@ pytest-xdist = "^2.3.0"
 
 [tool.poetry.extras]
 msgpack = ["msgpack"]
+numpy = ["numpy"]
 toml = ["toml"]
 yaml = ["pyyaml"]
-all = ["msgpack", "toml", "pyyaml"]
+all = ["msgpack", "toml", "pyyaml", "numpy"]
 
 [build-system]
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
+
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,13 +32,23 @@ jinja2 = "*"
 msgpack = { version = "*", markers = "extra == 'msgpack' or extra == 'all'", optional = true }
 toml = { version = "*", markers = "extra == 'toml' or extra == 'all'", optional = true }
 pyyaml = { version = "*", markers = "extra == 'yaml' or extra == 'all'", optional = true }
-numpy = { version = "*", markers = "extra == 'numpy' or extra == 'all'", optional = true }
+numpy = [
+    { version = "~1.21.0", markers = "python_version ~= '3.7.0' and (extra == 'numpy' or extra == 'all')", optional = true },
+    { version = ">1.21.0", markers = "python_version ~= '3.8.0' and (extra == 'numpy' or extra == 'all')", optional = true },
+    { version = ">1.21.0", markers = "python_version ~= '3.9.0' and (extra == 'numpy' or extra == 'all')", optional = true },
+    { version = ">1.22.0", markers = "python_version ~= '3.10' and (extra == 'numpy' or extra == 'all')", optional = true },
+]
 
 [tool.poetry.dev-dependencies]
 pyyaml = "*"
 toml = "*"
 msgpack = "*"
-numpy = "*"
+numpy = [
+    { version = "~1.21.0", markers = "python_version ~= '3.7.0'" },
+    { version = ">1.21.0", markers = "python_version ~= '3.8.0'" },
+    { version = ">1.21.0", markers = "python_version ~= '3.9.0'" },
+    { version = ">1.22.0", markers = "python_version ~= '3.10'" },
+]
 flake8 = "*"
 pytest = "*"
 pytest-cov = "*"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "pypy-3.8"]
+        # Remove pypy-3.8 until it supports numpy
+        python-version: ["3.7", "3.8", "3.9", "3.10"]  # "pypy-3.8"
         os: [ubuntu-20.04, macos-10.15, windows-2019]
     steps:
     - name: Checkout
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ venv/
 .idea/
 out/
 docs/CHANGELOG.md
+poetry.lock
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
       - id: end-of-file-fixer
 
   - repo: https://github.com/psf/black
-    rev: 19.3b0
+    rev: 22.3.0
     hooks:
       - id: black
         args: [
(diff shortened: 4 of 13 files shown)


yukinarit:
Hi @kmsquire! 

This PR is awesome! Let me review it on this weekend 👍

kmsquire:
Tests seem to be failing on 3.7.  Locally they pass for me on 3.9 on macos, but maybe I'm doing something that isn't compatible with 3.7.  I'll try to take a look.

kmsquire:
I think I fixed the issues on 3.7 (and 3.8).  The fix might be a little hacky, so I'm very open to changes.

yukinarit:
It appears Github Actions uses CPython 3.7 instead of pypy>=3.7. Let me fix it first and see if the poetry error disappears. 

> he currently activated Python version 3.6.12 is not supported by the project (^3.7.0).
Trying to find and use a compatible version. 
Using python3 (3.7.9)
Creating virtualenv pyserde--xklI_Zh-py3.7 in C:\Users\runneradmin\AppData\Local\pypoetry\Cache\virtualenvs
Installing dependencies from lock file

https://github.com/yukinarit/pyserde/runs/5701449127?check_suite_focus=true

yukinarit:
Looks like numpy 1.21.1 wheel is not provided for python 3.10? https://pypi.org/project/numpy/1.21.1/#files

@kmsquire 
Do you have any idea why the latest numpy 1.22.3 wasn't picked by poetry? 🤔

kmsquire:
> Do you have any idea why the latest numpy 1.22.3 wasn't picked by poetry? 🤔

It looks like numpy 1.22.3 only supports python 3.8-3.10 (not 3.7), so poetry probably picked a version that (it thought) supported >3.7.

I'll try to see if a conditional install will work (1.21.x on 3.7, 1.22.x on >=3.8).

kmsquire:
Okay, I fixed the numpy install issue.  

I hadn't run `make setup` previously, so I didn't have `pre-commit` installed.  I did this now and fixed linting issues.

`black` 19.3b0 is installed by pre-commit, but is broken right now because of an update to `click`.  Updating black to 22.3.0 fixes the issue, so that should probably be done in `.pre-commit-config.yaml`.

I think all that is remaining is to add some more tests to fill out test coverage (and whatever else you comment on).

kmsquire:
As a note to self when I have time, here is possible way to test the code run if `numpy` isn't installed:
https://stackoverflow.com/questions/2481511/mocking-importerror-in-python

poetry.lock
```diff
@@ -0,0 +1,1046 @@
+[[package]]
```
yukinarit, line 1: Could you exclude poetry.lock?

yukinarit (commented):
Hi, Thanks for the great PR! It looks almost good to me! 
Please check the review comments. Also, please format the code by `make fmt`

pyproject.toml
```diff
@@ -32,11 +32,16 @@ jinja2 = "*"
 msgpack = { version = "*", markers = "extra == 'msgpack' or extra == 'all'", optional = true }
 toml = { version = "*", markers = "extra == 'toml' or extra == 'all'", optional = true }
 pyyaml = { version = "*", markers = "extra == 'yaml' or extra == 'all'", optional = true }
+numpy = [
+    { version = "<1.22.0", markers = "python_version == '3.7' and (extra == 'numpy' or extra == 'all')", optional = true },
+    { version = "^1.22.0", markers = "(python_version ~= '3.8' or python_version ~= '3.9' or python_version ~= '3.10') and (extra == 'numpy' or extra == 'all')", optional = true },
```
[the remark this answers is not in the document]
kmsquire: I think I tried that and it didn't work, but I can check again.

poetry.lock
kmsquire: Yes, sorry.

kmsquire:
> Hi, Thanks for the great PR! It looks almost good to me! Please check the review comments. Also, please format the code by `make fmt`

Hi, to be clear: the formatting issue doesn't have anything to do with my code--I did run `make fmt`.  The problem is that the version of black that is being used depends on `click`, and a version of `click` was released that broke this (old) version of `black`.

The easiest solution to this problem would be to upgrade `black` to the latest version, which doesn't have this issue.  I can do that in this PR if you'd like, although it might be better as a separate PR.

poetry.lock
kmsquire: Fixed in latest (force) push

pyproject.toml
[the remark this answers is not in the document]
kmsquire: Yes, if I change the line to
```suggestion
    { version = "^1.22.0", markers = "python_version >= '3.8' and (extra == 'numpy' or extra == 'all')", optional = true },
```
then when running `poetry install` on python 3.7, it still tries to install numpy v 1.22.3 (which isn't compatible with python 3.7).  This is probably a poetry bug, but for now, the only way I found around it was to specify each version as in the line above.

kmsquire:
Okay, I removed `poetry.lock`, and I updated black to the latest version so that it could run properly in CI (and elsewhere).

As noted above, I did not change the numpy install conditions, because while your suggestion makes sense, it results in incorrect behavior on Python 3.7.

yukinarit:
CI tried to install numpy  1.22 for python 3.7

```
  Command ['/home/runner/.cache/pypoetry/virtualenvs/pyserde-tFPN8vZX-py3.7/bin/pip', 'install', '--no-deps', '/home/runner/.cache/pypoetry/artifacts/5b/d8/4a/ba9c80ec9b3f93ad0a31ca32f1fe044ff136c92d0e9a14fbe976d6d036/numpy-1.22.3.zip'] errored with the following return code 1, and output: 
```

How about updating the marker condition to use `~`? 🤔 

```
markers = "python_version ~= '3.7'
```

yukinarit:
@kmsquire Sorry for the delay. I don't mind upgrading black! 👍

kmsquire:
> CI tried to install numpy 1.22 for python 3.7

Strange, this was working before.  I'll take a look.

kmsquire:
> markers = "python_version ~= '3.7'

Unfortunately, this doesn't seem to work, nor has anything else I've tried locally. 😢 

I'm going to file a bug report with poetry.  

In the mean time, I'll back out the update to numpy, so we'll be on version 1.21.x for all versions of python.

kmsquire:
Getting closer (albeit with a restriction to numpy 1.21 for now).

The pypi test image needs to have some blas installed in order to compile numpy.  I'm not sure of the best way to do this, and probably won't be able to look into it further until sometime next week.

yukinarit:
It's quite hard work to setup build requirements for Windows, Linux and macOS. I don't mind removing `pypy` from CI build matrix.

kmsquire:
Okay, I think this should finally be good.  

The latest set of commits
* restricts numpy versions correctly based on python version
* removes pypy-3.8 from the CI test matrix (for now)

yukinarit:
@kmsquire it seems the CI error was introduced in   https://github.com/yukinarit/pyserde/pull/215 

Let me look into it 👍

yukinarit (approved):
LGTM!

yukinarit:
Closes #145

yukinarit:
@kmsquire Thank you for the great contribution! 🎉👍

PR #210, opened 2022-03-24
Add support for orjson, including direct numpy support

kmsquire:
This pull request builds on #209, adding support for orjson and modifying numpy support to pass values through to orjson when that module is chosen as the (de)serializer and the appropriate flags are passed.

Still needs tests and probably discussion around how much of orjson to pull in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Foo(i=10, s='foo', f=100.0, b=True)
 - Supported data formats
     - dict
     - tuple
-    - JSON
+    - JSON (via `json` or `orjson` packages)
 	- Yaml
 	- Toml
 	- MsgPack
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -51,6 +51,8 @@ Next, import pyserde (de)serialize API. For JSON:
 
 ```python
 from serde.json import from_json, to_json
+# OR:
+# from serde.orjson import from_json, to_json
 ```
 
 Similarly, you can use other data formats.
--- a/docs/supported-data-formats.md
+++ b/docs/supported-data-formats.md
@@ -50,6 +50,42 @@ Call `to_json` to serialize `Foo` object into JSON string and `from_json` to des
 Foo(i=10, s='foo', f=100.0, b=True)
 ```
 
+### Orjson
+
+Orjson is a fast alternative to the standard json library.  It is implemented as separate sub-module because it's behavior is slightly different than the standard library.
+
+```python
+>>> from serde.orjson import to_json, from_json
+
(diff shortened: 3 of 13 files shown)


kigawas:
Hi I think this can be superseded by #226

kmsquire:
> Hi I think this can be superseded by https://github.com/yukinarit/pyserde/pull/226

Hi thanks!  I didn't see your PR and actually updated this one today.  I just pushed the changes.

This takes a slightly different approach than yours.  Rather than replacing the default `pyserde.json` module, it instead creates a separate `pyserde.orjson` module, which is mostly a drop-in replacement.  I chose to do this for a couple of reasons:

1. Changing the return type of `to_json` to `bytes` instead of `str` is a breaking change, and a lot of code in the wild would have to be updated to support it.  It is possible to, e.g., assume `utf-8` and do that conversion in `pyserde`, but that also reduces flexibility in allowing the user to choose that conversion themselves (although I'm not sure how important that is).
2. Recent versions of `orjson` support passing dataclasses directly, which is faster than converting to a dictionary and then using orjson, but which doesn't support all of the features that `pyserde` offers.  In particular, it doesn't support `pyserde`'s `field` methods.  I added a manual fallback to use dictionary conversion if users need this functionality.

Either way, your changes are a bit simpler.  @yukinarit should take a look at both PRs and decide which is the better direction, or maybe if some of the ideas from each should be combined.

Cheers!

yukinarit:
@kmsquire @kigawas 
Thanks! Let me see both of the approaches get back to you! 🎉

kigawas:
Changing to `bytes` should not introduce significant breakage I think, since `json.loads` and `orjson.loads` both support `str` and `bytes`. 
Additionally, mixing `bytes` and `str` increases discrepancy rather than decreases, thus I prefer to unify serialized JSON with `bytes`.

yukinarit:
Sorry for the delay, I have been busy in the new environment. I finally got time to review 

The both of the PRs look great 👍 but kigawas's one is similar to what I was originally thinking, but I have the same concern as @kmsquire said.

Changing return type from bytes to str is a big breaking change for the existing users. I guess most of the Pythonistas would prefer/expect JSON library to produce the output in str not bytes. The implicit conversion from bytes to str sounds unnecessary overhead, but pyserde's design principle is the best simplicity than the best performance as opposed to orison which aims to (de)serialize with the least overhead. Also, If you don't the `decode` overhead, you can create a wrapper which calls `to_dict`? 🤔 

That being said,

@kigawas Is it ok to request to add `decode()`? 

@kigawas @kmsquire 
any feedback is appreciated 🙂

kmsquire:
Closing.  Related functionality merged as part of #226.

PR #213, opened 2022-04-03
fix: Add dataclass decorator for all example code

### LINKED ISSUE #205: mypy: Unexpected keyword argument
Hi,

This is very cool!

Unfortunately, when I use `mypy`, I get `Unexpected keyword argument` when I try to `return` a `@serde`ed class from a function. I'm not sure if `serde` would need to have a `mypy` plugin or if I'm making a mistake.

I can make a minimal test case if needed.

Thank you!

yukinarit:
@serde decorator works without @dataclass decorator but there is
no way to prevent mypy from producing type error. Adding @dataclass
decorator back to the all the example code.

Find related mypy page
https://mypy.readthedocs.io/en/stable/additional_features.html#caveats-known-issues

Closes #205

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Declare a class with pyserde's `@serde` decorator.
 
 ```python
 @serde
+@dataclass
 class Foo:
     i: int
     s: str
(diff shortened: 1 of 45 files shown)


gitpushdashf:
This looks good to me!

PR #222, opened 2022-04-28
Numpy updates

kmsquire:
Here are a few fixes/updates for numpy support

* Add support for serializing numpy.datetime64 objects
* Allow direct numpy array deserialization
* Only define np_get_origin() for python 3.8 or earlier
* Fill out tests


kmsquire:
The code coverage differences are expected, since most (maybe all) of the missing lines are in code that only runs on certain versions of python (e.g., <=3.8), or without numpy, etc.

The benchmark failure is known.

yukinarit (approved):
LGTM!

PR #238, opened 2022-06-13
feat: Don't wrap user exception in SerdeError

### LINKED ISSUE #233: Raising an exception in `__post_init__` results in `SerdeError`?
I have a dataclass and I want do to some validation of the fields in `__post_init__`. If there is a problem I raise a custom exception. However, `pyserde` catches all exceptions in `from_obj`and raises a `SerdeError` instead? This isn't really what I want and I don't know if it if the best behavior in general?

yukinarit:
Closes #233 

@AustinScola
PTAL.

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -12,7 +12,7 @@ import pathlib
 import re
 import uuid
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, Iterator, List, Mapping, Optional, Type, TypeVar, Union
+from typing import Any, Callable, Dict, List, Mapping, Optional, Type, TypeVar, Union
 
 import stringcase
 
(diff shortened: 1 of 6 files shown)


AustinScola (commented):
Nice, thanks for taking a look at this. Could you help me understand where the exception gets wrapped into a `UserError`?

serde/core.py
```diff
@@ -502,11 +502,11 @@ def supports_default(self) -> bool:
 F = TypeVar('F', bound=Field)
 
 
-def fields(field_cls: Type[F], cls: Type) -> Iterator[F]:
+def fields(field_cls: Type[F], cls: Type) -> List[F]:
```
AustinScola, line 505: Does this have to be a list for the templating to work?

yukinarit: Yeah, it was used twice in the template. Returning `List` is more convenient.  Thanks for reviewing!

PR #251, opened 2022-07-23
Use str instead of repr in typename, fixing Union[Literal["str"], ...]

kngwyu:
Hi, thank you for this package 😃 

- `str` is more similar to the behavior of format, which is used elsewhere in the function
-  this fixes `Union[something, Literal["str]]`
  - the current implementation tries to use something like `def union_de_something_literal_'str'` as a function name, which contains `'` and thus is invalid

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -3,14 +3,9 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import Dict, Generic, List, NewType, Optional, Set, Tuple, TypeVar, Union
 
-if sys.version_info[:2] == (3, 7):
-    from typing_extensions import Literal
-else:
-    from typing import Literal
-
-
 import serde
 from serde.compat import (
+    Literal,
     is_dict,
     is_generic,
     is_list,
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -3,6 +3,12 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import Dict, Generic, List, NewType, Optional, Set, Tuple, TypeVar, Union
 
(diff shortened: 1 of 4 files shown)


kngwyu:
I added another commit to support `Union[Literal["*"], ...]`

tests/test_compat.py
```diff
@@ -3,6 +3,12 @@
 from datetime import datetime
 from typing import Dict, Generic, List, NewType, Optional, Set, Tuple, TypeVar, Union
 
+if sys.version_info[:2] == (3, 7):
+    from typing_extensions import Literal
+else:
+    from typing import Literal
```
yukinarit, line 9: You can import `Literal` from `serde.compat`  👍 

https://github.com/yukinarit/pyserde/blob/master/serde/compat.py#L21-L26

tests/test_union.py
```diff
@@ -2,7 +2,7 @@
 import sys
 from dataclasses import dataclass
 from ipaddress import IPv4Address
-from typing import Dict, Generic, List, Optional, Tuple, TypeVar, Union
+from typing import Dict, Generic, List, Literal, Optional, Tuple, TypeVar, Union
```
yukinarit, line 5: You can import `Literal` from `serde.compat`  👍 

https://github.com/yukinarit/pyserde/blob/master/serde/compat.py#L21-L26

yukinarit (commented):
Hi @kngwyu!

Thank you for your contribution. I left a few comments, PTAL 👍

kngwyu:
Thanks, I fixed imports

yukinarit (approved):
LGTM 👍

PR #252, opened 2022-07-25
Support NumPy 1.23 in Python 3.7/3.8

### LINKED ISSUE #246: Numpy v1.23.0 breaks pyserde compatability in Python versions <= 3.8
The commit https://github.com/numpy/numpy/commit/7739583f5fb39c31e83010a3153fa078004e55eb splits the typing library into a public and private part.

A consequence is that pyserde's compatibility mode for Python 3.8 (and below) which relies on some private types will no longer work as those types have now moved.

kngwyu:
Fix #246 

I didn't test it locally since I only have Python 3.10..., but let's see what happens 😅


yukinarit (approved):
LGTM, Thanks you!

PR #254, opened 2022-07-27
Support native datetime types in toml and yaml

kngwyu:
toml and yaml support native datetime types (see e.g., https://toml.io/en/v1.0.0#offset-date-time).
Currently pyserde converts all datetime types to string, which prevents tomli and pyyaml from handling these types directly.


kngwyu:
BTW msgpack doesn't support datetime https://github.com/msgpack/msgpack-python/issues/252

serde/se.py
```diff
@@ -291,7 +292,15 @@ def is_serializable(instance_or_class: Any) -> bool:
...
 
 
-def to_obj(o, named: bool, reuse_instances: bool, convert_sets: bool, c: Type = None, type_check: TypeCheck = Coerce):
+def to_obj(
+    o,
+    named: bool,
+    reuse_instances: bool,
+    convert_sets: bool,
+    c: Type = None,
+    preserved_datetime_types: Tuple[type, ...] = (),
```
yukinarit, line 301: Hi @kngwyu 

Thank you for the PR! It's a good idea 👍 

Passing the tuple of types here in the function means you want to use a different (de)serializer depending on the type? (e.g Use ISO for `datetime` and RFC3339 for `date`)? 🤔 

Another idea is to allow a user to pass their own (de)serializer for datetime types, 

i.e.

#### Serialization

```python
def to_obj(
    ...
    # pyserde preserve datetime objects if option is None
    datetime_serializer: Optional[Callable] = lambda dt: dt.isoformat(),
    ...
```

#### Deserialization
```python
def from_obj(
    ...
    # pyserde preserve datetime objects if option is None
    datetime_deserializer: Optional[Callable] = lambda typ, v: typ.fromisoformat(v),
    ...
):
```

NOTE: The above code wasn't tested.

What do you think? 🤔

kngwyu: My intention is to handle the three file formats below:
1. toml supports date, time, and datetime
2. yaml supports date and datetime, but it doesn't support local time without date
3. json doesn't have relevant types, so all datetime types are often treated as ISO8601 string

(Note that msgpack supports timestamp but it's not fully compatible with datetime types.)
Of course, we can make it more flexible by accepting callbacks, but because I expect most users want to use the library *as is* and don't want to customize the behavior heavily, I don't feel an urgent need for that.

kngwyu: In other words, I think that the behavior should be file-format dependent, and it doesn't have to be such flexible for users. Users may not know what types are supported by yaml/toml/etc.

yukinarit: OK, now I get it.  (Sorry for the delay in getting back to you)

Do you think it's feasible/reasonable to implement a more generalized `preseve_types` not just for datetime types but also for like `Set` or `Decimal`? 🤔  because the parameter `converts_set` is very similar to `preserved_datetime_types`

kngwyu: > Do you think it's feasible/reasonable to implement a more generalized preseve_types not just for datetime types but also for like Set or Decimal? thinking because the parameter converts_set is very similar to preserved_datetime_types

I think so, and it sounds like a good idea because I feel like these internal (generated) functions should have fewer arguments for simplicity. I'll try that in my spare time, possibly a few days later.

> Sorry for the delay in getting back to you

No pressures. Thank you for your comment!

kngwyu:
I'm sorry for doing nothing for a long time 😅 
Closing this now because I guess it's outdated, but I'll re-implement the fix if the datetime issue is still there

yukinarit:
Hi @kngwyu 

Thanks for your message. I believe `reuse_instances` resolves this issue. https://github.com/yukinarit/pyserde/issues/483 :+1:

PR #284, opened 2022-11-26
Pickle serializer/deserializer implementation.

DoeringChristian:
I have copied the yaml serializer and deserializer implementation and adapted it to work with pickle.
I'm not quite sure if this will find uses but I think it is nice to have direct serialization/deserialisation to pickle.
Added from_pickle and to_pickle methods to the tests.


yukinarit (approved):
LGTM, Thank you!

PR #294, opened 2022-12-14
Add support for optional unions

### LINKED ISSUE #293: Optional Unions Don't Work?
See the following minimal example:

```python
from serde import yaml, serde
from typing import Optional, Union

@serde
@dataclass(frozen=True)
class Foo:
    request: Optional[str]

@serde
@dataclass(frozen=True)
class Bar:
    request: Optional[Union[str, int]]


print(yaml.from_yaml(Foo, "{}"))
print(yaml.from_yaml(Bar, "{}"))

```

Expectation is that the second `print` results in `Bar(request=None)`, but instead raises `serde.compat.SerdeError: 'request'`.

Tested on python 3.11.0 and 3.9.7. pyserde version is 0.9.6.

Full stack trace:

```
Foo(request=None)
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.9/site-packages/serde/de.py", line 354, in from_obj
    res = serde_scope.funcs[func_name](c, maybe_generic=maybe_generic, data=o, reuse_instances=reuse_instances)
  File "<string>", line 11, in from_dict
KeyError: 'request'

During handling of the above exception, another exception occurred:

  File "/Users/stuart/tmp/serde-union-fail.py", line 18, in <module>
    print(yaml.from_yaml(Bar, "{}"))
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.9/site-packages/serde/yaml.py", line 46, in from_yaml
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.9/site-packages/serde/de.py", line 441, in from_dict
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.9/site-packages/serde/de.py", line 418, in from_obj
    raise SerdeError(e)
serde.compat.SerdeError: 'request'
```

soaxelbrooke:
Fixes  #293.

Adds a check for optionals in union types (and optional unions), allowing parsing when the optional field is missing.

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -9,7 +9,7 @@ import dataclasses
 import functools
 import typing
 from dataclasses import dataclass, is_dataclass
-from typing import Any, Callable, Dict, List, Optional, TypeVar
+from typing import Any, Callable, Dict, List, Optional, TypeVar, get_args
 
 import jinja2
 from typing_extensions import Type, dataclass_transform
@@ -515,6 +515,8 @@ class DeField(Field):
     def data(self) -> str:
         if self.iterbased:
             return f'{self.datavar}[{self.index}]'
+        elif is_union(self.type) and type(None) in get_args(self.type):
+            return f'{self.datavar}.get("{self.conv_name()}")'
         else:
             return f'{self.datavar}["{self.conv_name()}"]'
 
--- a/serde/de.py
+++ b/serde/de.py
@@ -12,12 +12,6 @@ import typing
 from dataclasses import dataclass, is_dataclass
 from typing import Any, Callable, Dict, List, Optional, TypeVar
 
-if sys.version_info[1] < 8:
-    # get_args was added in python 3.8
-    from typing_utils import get_args
-else:
-    from typing import get_args
-
 import jinja2
 from typing_extensions import Type, dataclass_transform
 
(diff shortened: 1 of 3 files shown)


soaxelbrooke:
@yukinarit This coverage regression is because codecov only considers one test run, thus doesn't see that both branches are tested in the relevant python version. Should be fine to override?

serde/de.py
```diff
@@ -7,10 +7,17 @@
...
+import sys
 import typing
 from dataclasses import dataclass, is_dataclass
 from typing import Any, Callable, Dict, List, Optional, TypeVar
 
+if sys.version_info[1] < 8:
+    # get_args was added in python 3.8
+    from typing_utils import get_args
+else:
+    from typing import get_args
```
yukinarit, line 19: Can we use [`compat.get_args`](https://github.com/yukinarit/pyserde/blob/648471c85398d33a4d8627462de4db17094ef639/serde/compat.py#L130-L137) instead? 🤔 

I don't really know the differences between typing_extensions and typing_utils though

yukinarit (commented):
Hi @soaxelbrooke! Thank you so much for the PR! 

Left a comment. PTAL

serde/de.py
soaxelbrooke: Thanks, didn't know about this! Updated.

yukinarit (approved):
LGTM

Issue #309, opened 2023-02-17
Serialize dataclasses without `serde` decorator


Kobzol:
Hi, it's me again :grin: Sorry for spamming so many issues, but `pyserde` is simply the best serialization library for Python that I have encountered yet, so I'm now migrating code to it and thus have a lot of questions :D

I wonder if it would be possible to implement support for serializing fields with a dataclass type, even if that dataclass isn't marked with `@serde`. I suppose that it could work by dynamically generating the serializing/deserializing methods when the type is first encountered and then caching them. Since the decorator is not used on such dataclasses, it would behave like `@serde`, without any parameters.
Basically, if `pyserde` saw a dataclass without the serde scope attribute, it could dynamically add it to the dataclass (or create an out-of-band hashmap of such dataclass types, to avoid modifying them), generate the serialization/deserialization functions and then continue as normal.

I know that (Rust) `serde` can't really do this, but in Python this should be theoretically possible. I have two use-cases for this:
1) Some type analysers do not yet support the `@dataclass_transform` attribute and if I add `@serde` to a dataclass, they stop all type checking of the dataclass :( I encounter this with PyCharm, `pyright` works fine. I hope that this will get resolved in time, but it's still a complication for me.
2) It would enable serialization of dataclasses that cannot be decorated with serde, e.g. ones from external code. Currently it can be resolved e.g. with a wrapper dataclass and `flatten=True`, but it would be nice if it was supported directly.

    Even for dataclasses from my own code, most of them are just marked with `@serde` without any additional configuration, but I still have to sprinkle this serialization decorator everywhere. It would be nice if such dataclasses (that do not require any specialized serde options) could be serialized automatically, without them even knowing about serde.

Let me know what you think about this! I can also help with implementing this or some of the other issues if you can guide me a bit.

yukinarit:
Hi @Kobzol 

Yes, I think this is possible. If you're interested in contributing, I would love to assist you 👍

PR #312, opened 2023-02-20
Generate serialization/deserialization code for dataclasses not marked with @serde

Kobzol:
This is an attempt to resolve https://github.com/yukinarit/pyserde/issues/309. It generates serialization/deserialization code for a nested dataclass field if that field type is not marked with `@serde`. Currently, the ser/de code is generated based on the `@serde` configuration of the containing dataclass, which is not ideal, it should probably be generated with an "empty" `@serde` configuration. But for that I'd need to store the type stack (which is used as a recursion guard) outside the `serialize` function as a global variable (?).

Let me know what you think about this approach.

Fixes: https://github.com/yukinarit/pyserde/issues/309

### DIFF
--- a/tests/test_basics.py
+++ b/tests/test_basics.py
@@ -1024,3 +1024,19 @@ def test_class_var() -> None:
     f = Nested()
     assert {"v": {"v": 100}} == serde.to_dict(f)
     assert f == serde.from_dict(Nested, {})
+
+
+def test_nested_dataclass_without_serde() -> None:
+    @dataclasses.dataclass
+    class Foo:
+        v: int
+
+    @serde.serde
+    @dataclasses.dataclass
+    class Wrapper:
+        foo: Foo
+
+    a = Wrapper(foo=Foo(v=1))
+    serialized = serde.to_dict(a)
+    assert {'foo': {'v': 1}} == serialized
+    assert a == serde.from_dict(Wrapper, serialized)
--- a/serde/de.py
+++ b/serde/de.py
@@ -213,7 +213,13 @@ def deserialize(
     Foo(i=10, dt=datetime.datetime(2021, 1, 1, 0, 0))
     """
 
+    stack = []
+
     def wrap(cls: Type):
+        if cls in stack:
+            return
+        stack.append(cls)
+
         tagging.check()
 
         # If no `dataclass` found in the class, dataclassify it automatically.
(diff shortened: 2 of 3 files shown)


tests/test_basics.py
```diff
@@ -1024,3 +1024,19 @@ class Nested:
...
+
+def test_nested_dataclass_without_serde() -> None:
+    @dataclasses.dataclass
+    class Foo:
+        v: int
+
+    @serde.serde
+    @dataclasses.dataclass
+    class Wrapper:
+        foo: Foo
```
yukinarit, line 1037: Do you also want to serialize/deserialize `Wrapper` without `@serde`? 🤔 
Maybe it's possible if we checks and generate code in [from_obj](https://github.com/yukinarit/pyserde/blob/2582e244642dc1ba1df68f7ff04018b018db735c/serde/de.py#L352) and [to_obj](https://github.com/yukinarit/pyserde/blob/2582e244642dc1ba1df68f7ff04018b018db735c/serde/se.py#L323)? What do you think?

yukinarit (commented):
Thank you for the PR! I left a comment, PTAL 🙏

tests/test_basics.py
Kobzol: I was also thinking about this. It already sort-of works for serialization of dataclasses, although it just uses `dataclasses.asdict`, so it only works for basic stuff.

Not having `@serde` on the "root" type is not that important to me specifically, but I think that it should be supported for consistency. It seems slightly orthogonal though, do you want me to implement it in this PR or in a follow-up one?

yukinarit: I am ok with the follow-up PR!

Kobzol:
Oh, I still think that we should resolve the `serde` configuration of the generated code. I'll send another PR for it.

Because currently, if you do something like this:
```python
@dataclass
class Foo:
    ...

@serde(foo=bar)
class W1:
    foo: Foo


@serde
class W2:
    foo: Foo
```

`Foo` will probably be generated with `foo=bar`, but if you switch the order of the wrappers, it will be generated without any options. That seems quite unintuitive and it should probably be generated without any `serde` options, using the defaults.

yukinarit:
@Kobzol Ah ok, thanks for the heads up 👍

PR #331, opened 2023-03-24
fix: Fixed broken strict type checking for Optionals. Fixes #330

### LINKED ISSUE #330: Strict type checking not working for Optional
Noticed that strict type checking does not seem to work properly for Optional[T]
For Example:

```python
import serde
from typing import Optional


@serde.serde(type_check=serde.Strict)
@dataclass
class Struct:
    value: int
    some: Optional[int]

print(serde.from_dict(Struct, {"value": 123, "some": "123"}))
```
returns Struct(value=123, some="123") when it should fail the type check for the inner type.

Fredrik-Reinholdsen:
Single addition to core that fixes strict type checking for Optional types.
See #330 for example.

All tests pass.

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -722,6 +722,7 @@ def {{type_check_func}}(self):
            compat.is_list(f.type) or
            compat.is_dict(f.type) or
            compat.is_tuple(f.type) or
+           compat.is_opt(f.type) or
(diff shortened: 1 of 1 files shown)


yukinarit (approved):
LGTM, Thank you!

PR #335, opened 2023-03-27
Serialize complex numbers

robertodr:


robertodr:
I think I need some guidance to implement this. I can't quite figure out where the appropriate point to unpack $a + \mathrm{i} b$ into a `(a, b)` tuple is. And also, once I have the tuple how to deserialize it back into $a+\mathrm{i} b$ format.

yukinarit:
Hi @robertodr, Thank you for your effort! 🙏 

Please try implementing render function for `complex` type for both serialize and deserialize. Render function renders the code in string, which is passed in `exec` later to generate real python functions.

* [serialize](https://github.com/yukinarit/pyserde/blob/2c24ce0f7c6075519cf9c5b359a161e41ff3125b/serde/se.py#L691-L736)
* [deserialize](https://github.com/yukinarit/pyserde/blob/2c24ce0f7c6075519cf9c5b359a161e41ff3125b/serde/de.py#L614-L662)

Please take a look at [this PR](https://github.com/yukinarit/pyserde/pull/286/files) that I made to support DefaultDict.

PR #354, opened 2023-05-24
Fix combination of alias and default

### LINKED ISSUE #353: Aliases are not checked when a default value is provided
When providing a default value, pyserde doesn't check any aliases.

Example:
```python

from serde import field, serde
from serde.json import from_json


@serde
@dataclass
class Foo:
    a: int = field(default=2, alias=["b", "c", "d"])


def main():
    s = '{"a": 10}'
    # Expected: From Json: Foo(a=10)
    # Actual: From Json: Foo(a=10)

    s = '{"b": 20}'
    # Expected: From Json: Foo(a=20)
    # Actual: From Json: Foo(a=2)


if __name__ == "__main__":
    main()
```

kmsquire:
Fixes #353

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -123,6 +123,13 @@ def _get_by_aliases(d: Dict[str, str], aliases: List[str]):
         return _get_by_aliases(d, aliases[1:])
 
 
+def _exists_by_aliases(d: Dict[str, str], aliases: List[str]):
+    for alias in aliases:
+        if alias in d:
+            return True
+    return False
+
+
 def _make_deserialize(
     cls_name: str,
     fields,
@@ -262,6 +269,7 @@ def deserialize(
         g['NoCheck'] = NoCheck
         g['coerce'] = coerce
         g['_get_by_aliases'] = _get_by_aliases
+        g['_exists_by_aliases'] = _exists_by_aliases
         if deserialize:
             g['serde_custom_class_deserializer'] = functools.partial(
                 serde_custom_class_deserializer, custom=deserializer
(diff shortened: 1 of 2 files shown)


serde/de.py
```diff
@@ -876,8 +884,12 @@ def literal(self, arg: DeField) -> str:
         func_name = literal_func_name(type_args(arg.type))
         return f"serde_scope.funcs['{func_name}'](cls=cls, data={arg.data}, reuse_instances=reuse_instances)"
 
+    # HERE
```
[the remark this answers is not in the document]
kmsquire: Nope, sorry, will remove (and rebase).

kmsquire:
Rebased and removed the spurious comment.

kmsquire:
Added a couple of lines to the test code to fix code coverage.  Should be good to go.

serde/de.py
```diff
@@ -878,6 +886,9 @@ def literal(self, arg: DeField) -> str:
 
     def default(self, arg: DeField, code: str) -> str:
         exists = f'"{arg.conv_name()}" in {arg.datavar}'
+        if arg.alias:
+            aliases = map(lambda s: f'"{s}"', [arg.name, *arg.alias])
+            exists += f' or _exists_by_aliases({arg.datavar}, [{",".join(aliases)}])'
```
yukinarit, line 891: I saw the generated code for the following `Foo` class. 

```python
    @serde.serde
    class Foo:
        a: int = serde.field(default=2, alias=["b", "c", "d"])
```

Here is the code excerpt from `from_dict`.
```python
    __a = (
        (_get_by_aliases(data, ["a", "b", "c", "d"]))
        if "a" in data or _exists_by_aliases(data, ["a", "b", "c", "d"])
        else serde_scope.defaults["a"]
    )
```

I think "a" is checked twice? Maybe we can further optimize to generate the code like this? 🤔  
```python
    __a = (
        (_get_by_aliases(data, ["a", "b", "c", "d"]))
        _exists_by_aliases(data, ["a", "b", "c", "d"])
        else serde_scope.defaults["a"]
    )
```

yukinarit: I tested this works 👍 
```python
    def default(self, arg: DeField, code: str) -> str:
        if arg.alias:
            aliases = map(lambda s: f'"{s}"', [arg.name, *arg.alias])
            exists = f'_exists_by_aliases({arg.datavar}, [{",".join(aliases)}])'
        else:
            exists = f'"{arg.conv_name()}" in {arg.datavar}'

```

yukinarit (commented):
Thanks, I found 1 more thing, Please check!

serde/de.py
kmsquire: Thanks!  I pushed your version.

PR #377, opened 2023-06-08
Fix nested generic class deserialization

### LINKED ISSUE #376: Incorrect handling of nested generic classes
`pyserde` does not correctly identify the generic type(s) for generic classes nested within another generic class when there is more than one inner generic class.

For example, in the example below, when deserializing `Foo12`, `pyserde` thinks that the generic type for `Baz[U]` is `Baz[Enum1]`, when it is in fact `Baz[Enum2]`:

```python
from enum import Enum
from typing import Generic, TypeVar

import serde
from serde import from_dict, to_dict

T = TypeVar("T")
U = TypeVar("U")

@serde.serde
class Bar(Generic[T]):
    v: T


@serde.serde
class Baz(Generic[U]):
    y: U


@serde.serde
class Foo(Generic[T, U]):
    bar: Bar[T]
    baz: Baz[U]


class Enum1(Enum):
    A = "a"
    B = "b"

class Enum2(Enum):
    C = "c"
    D = "d"

Bar1 = Bar[Enum1]
Baz2 = Baz[Enum2]

Foo12 = Foo[Enum1, Enum2]

foo12 = Foo12(Bar1(Enum1.A), Baz2(Enum2.C))

d = to_dict(foo12)
print(d)

e = from_dict(Foo12, d)  # Fails with "<Enum2.C: 'c'> is not a valid Enum1"
print(e)
```

kmsquire:
During deserialization of a generic class with variables that also had generic class types, we were incorrectly looking up the concrete class name, by simply checking the position of a type variable in the inner class.

Instead, we now use the position of the type variable to get the name of the corresponding type variable in the parent class, and then find the position of that variable in the list of type variables used when defining the parent class.

For example, consider these definitions:

```python
@serde
class Bar[T]:
    t: T

@serde
class Foo[U,V]
    u: U
    bar: Bar[V]

foo = Foo[int, str](1, Bar[str]('a'))

from_json(to_json(foo))
```

When deserializing, we
1. get the index of `T` in the list of `Bar`'s type vars (position 0) 
2. use that to find the name of `Bar`'s type variable in `Foo` (i.e., `V`)
3. find the index of `V` in `Foo`'s type variable list (position 1)
4. finally use that index to determine the concrete type of `V` from `Foo[int, str]` (i.e., `str`)

(It would be nice if we could get the `str` type directly from `Bar[str]` on the line defining `foo`, but this does not seem to be possible currently.)

The fix is a little complicated, so any suggestions on a simpler fix (or anything, really) would be welcome.

Fixes #376

### DIFF
--- a/serde/compat.py
+++ b/serde/compat.py
@@ -812,6 +812,28 @@ def is_ellipsis(typ: Any) -> bool:
     return typ is Ellipsis
 
 
+def get_type_var_names(cls: Type[Any]) -> Optional[List[str]]:
+    """
+    Get type argument names of a generic class.
+
+    >>> T = typing.TypeVar('T')
+    >>> class GenericFoo(typing.Generic[T]):
+    ...     pass
+    >>> get_type_var_names(GenericFoo)
+    ['T']
+    >>> get_type_var_names(int)
+    """
+    bases = getattr(cls, "__orig_bases__", ())
+    if not bases:
+        return None
+
+    type_arg_names: List[str] = []
+    for base in bases:
+        type_arg_names.extend(arg.__name__ for arg in get_args(base))
+
+    return type_arg_names
+
+
 def find_generic_arg(cls: Type[Any], field: TypeVar) -> int:
     """
     Find a type in generic parameters.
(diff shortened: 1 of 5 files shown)


kmsquire:
FWIW, I just looked, and I don't quite see how I can increase the patch code coverage easily.

yukinarit:
@kmsquire Thank you so much!
Let me review this tomorrow 👍

yukinarit (commented):
@kmsquire Mostly LGTM 🎉 
I left small nitpick comments, PTAL.

yukinarit:
FYI: I also tested a more nested case. It works.

```python
from typing import Generic, TypeVar

import serde
from serde import from_dict, to_dict

T = TypeVar("T")
U = TypeVar("U")
V = TypeVar("V")


@serde.serde
class Bar(Generic[T, V]):
    a: T
    b: V


@serde.serde
class Baz(Generic[U]):
    c: U


@serde.serde
class Qux(Generic[V]):
    d: V


@serde.serde
class Foo(Generic[T, U, V]):
    bar: Bar[T, V]
    baz: Baz[U]


BarStr = Bar[str, float]
BazInt = Baz[int]

FooStrInt = Foo[str, int, Qux[float]]

f = FooStrInt(Bar("hello", Qux(100.0)), Baz(10))

print("original", f)

d = to_dict(f)
print("serialized", d)

ff = from_dict(FooStrInt, d)
print("deserialized", ff)
```

kmsquire:
> FYI: I also tested a more nested case. It works.

Also, FWIW, I think this will still fail if you change the definition of `Qux` to be

```python
@serde.serde
class Qux(Generic[W]):
    d: W
```

Or it might need the second generic variable of `Bar` to change as well--I'm not sure.  But I think if the generic variable names are not the same for a nested variable two levels down, the code won't find the correct mapping.

Unfortunately, I can't test this right now, and probably won't be able to for the next couple of weeks.  Also, assuming this actually fails, I also don't really have a solution.  That said, this fix should handle most cases.

yukinarit:
@kmsquire Yes, I understand 👍

yukinarit (approved):
LGTM 🚀

yukinarit:
> FWIW, I just looked, and I don't quite see how I can increase the patch code coverage easily.

I think it's because of the `typing.overload` is uncovered. I will try [the suggested workaround](https://github.com/nedbat/coveragepy/issues/970#issuecomment-612602180).

kmsquire:
FWIW, I just realized that I started a "review" and responded to most of your comments, but then forgot to send the responses.  Sorry!  Anyway, thanks for merging.  If I think of anything else that might help with the edge case, I'll let you know (but I don't think so).

PR #378, opened 2023-06-11
Fix variable length tuple codegen

kykosic:
This fixes an issue where if you have a variable length tuple whose elements are `dataclass`, it doesn't properly codegen tuple-parsing logic for the elements. 

Example:
```python
from serde import serde, to_dict

@serde
@dataclass
class Foo:
    a: int
    
@serde
@dataclass
class Bar:
    foos: tuple[Foo, ...]
    
bar = Bar(foos=(Foo(0), Foo(1)))
to_dict(bar)
```
```
# old output
{'foos': (Foo(a=0), Foo(a=1))}

# new output
{'foos': ({'a': 0}, {'a': 1})}
```

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -9,7 +9,7 @@ import dataclasses
 import functools
 import typing
 from dataclasses import dataclass, is_dataclass
-from typing import Any, Callable, Dict, List, Optional, TypeVar, Generic, overload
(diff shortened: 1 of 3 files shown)


yukinarit (approved):
LGTM, thank you!

PR #412, opened 2023-07-22
Fix frozen dataclass inheritance

yukinarit:
Fix frozen dataclass inheritance

`should_impl_dataclass` should not compare the length of declared
fields as `dataclasses.field` might include inherited fields but
`__annotations__` does not.

This commit allows to declare frozen dataclass to inherit from
another frozen dataclass.

```python
@serde
@dataclass(frozen=True)
class Foo:
    a: int

@serde
@dataclass(frozen=True)
class Bar(Foo):
    b: int
```

However, there is still an issue. It seems type checkers e.g. mypy
and pyright can not correctly understand the frozen pyserde class
due to the limitation of `dataclass_transform`. You will get a type
error "A frozen class cannot inherit from a class that is not frozen"

https://peps.python.org/pep-0681/#dataclass-semantics

The PEP681 says
> A class that has been decorated with dataclass_transform is
considered neither frozen nor non-frozen

If you are mypy/pyright user, I'd suggest to declare dataclass
without serde decorator as a workaround. This way won't produce
any type errors.

```python
@dataclass(frozen=True)
class Foo:
    a: int

@dataclass(frozen=True)
class Bar(Foo):
    b: int
```

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -881,9 +881,6 @@ def should_impl_dataclass(cls: Type[Any]) -> bool:
     if not annotations:
         return False
 
-    if len(annotations) != len(dataclasses.fields(cls)):
-        return True
-
     field_names = [field.name for field in dataclass_fields(cls)]
     for field_name in annotations:
         if field_name not in field_names:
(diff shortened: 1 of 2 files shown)


AustinScola:
You might want to look into creating a mypy plugin so mypy understands the `serde` decorator?

Issue #297, opened 2023-01-17
kw_only not supported


jennydaman:
```python
from serde import deserialize, from_dict
from dataclasses import dataclass


@deserialize
@dataclass(kw_only=True)
class Hello:
    a: str


assert Hello(a='ok') == from_dict(Hello, {'a': 'ok'})
```

```
Traceback (most recent call last):
  File "/home/jenni/fnndsc/pythonProject/venv/lib/python3.10/site-packages/serde/de.py", line 356, in from_obj
    res = serde_scope.funcs[func_name](c, maybe_generic=maybe_generic, data=o, reuse_instances=reuse_instances)
  File "<string>", line 21, in from_dict
serde.compat.UserError: Hello.__init__() takes 1 positional argument but 2 were given

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jenni/fnndsc/pythonProject/main.py", line 11, in <module>
    assert Hello(a='ok') == from_dict(Hello, {'a': 'ok'})
  File "/home/jenni/fnndsc/pythonProject/venv/lib/python3.10/site-packages/serde/de.py", line 443, in from_dict
    return from_obj(cls, o, named=True, reuse_instances=reuse_instances)
  File "/home/jenni/fnndsc/pythonProject/venv/lib/python3.10/site-packages/serde/de.py", line 417, in from_obj
    raise e.inner
  File "<string>", line 15, in from_dict
TypeError: Hello.__init__() takes 1 positional argument but 2 were given
```

Known versions affected: 0.8.3, 0.9.7

yukinarit:
@jennydaman supported in v0.12.2

PR #413, opened 2023-07-22
Add support for dataclasses with `kw_only`

m472:
Python 3.10 introduces the `kw_only` argument to the dataclass decorator. This is not compatible with current versions of pyserde (see #297 ).

The `kw_only` argument can be useful for dataclasses making use of inheritance because the "TypeError: non-default argument 'child_val' follows default argument" error.

The change in this pull request changes the `from_dict` method to always use named arguments. Because there are no positional only arguments in dataclasses (at least to my knowledge) there should not be any issue with this.

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -627,6 +627,7 @@ class Field(Generic[T]):
             deserializer=deserializer,
             flatten=flatten,
             parent=parent,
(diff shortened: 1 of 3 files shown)


yukinarit (approved):
LGTM, Thank you!

Issue #437, opened 2023-10-21
SerdeError when serializing a frozenset that is part of a union field


acolley-gel:
Hello,

I came across this issue that seems to present when a frozenset is part of a union, either directly as part of the union or as a nested dataclass with a field that is a frozenset:

```python
from dataclasses import dataclass

from serde import serde
from serde.se import to_dict


@serde
@dataclass
class WithSet:
    stuff: set[int]


@serde
@dataclass
class WithFrozenSet:
    stuff: frozenset[int]


@serde
@dataclass
class Working1:
    field: WithFrozenSet


@serde
@dataclass
class Working2:
    field: WithSet | int


@serde
@dataclass
class Working3:
    field: frozenset[int]


@serde
@dataclass
class NotWorking1:
    field: WithFrozenSet | int


@serde
@dataclass
class NotWorking2:
    field: frozenset[int] | int


to_dict(Working1(WithFrozenSet(frozenset({1}))))
to_dict(Working2(WithSet({1})))
to_dict(Working2(1))
to_dict(Working3(frozenset({1})))

to_dict(NotWorking1(WithFrozenSet(frozenset({1}))))

to_dict(NotWorking2(frozenset({1})))
```

```
serde.compat.SerdeError: Can not serialize WithFrozenSet(stuff=frozenset({1})) of type WithFrozenSet for Union[WithFrozenSet, int]
```

```
serde.compat.SerdeError: Can not serialize frozenset({1}) of type Set for Union[Set[int], int]
```

The scenario I encountered is similar to `NotWorking1`.

This issue seems related to #427.

Python: 3.12.0
pyserde: 0.12.3

yukinarit:
@acolley-gel v0.12.4 was pushed to PyPI!

PR #438, opened 2023-10-23
Fix frozenset union with prim #437

acolley-gel:
Proposed fix for #437

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -413,7 +413,7 @@ def is_list_instance(obj: Any, typ: Type[Any]) -> bool:
 
 
 def is_set_instance(obj: Any, typ: Type[Any]) -> bool:
-    if not isinstance(obj, set):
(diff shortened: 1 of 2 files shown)


serde/core.py
```diff
@@ -413,7 +413,7 @@ def is_list_instance(obj: Any, typ: Type[Any]) -> bool:
 
 
 def is_set_instance(obj: Any, typ: Type[Any]) -> bool:
-    if not isinstance(obj, set):
+    if not isinstance(obj, (set, frozenset)):
```
yukinarit, line 416: ah, fronzenset is not derived from set. Interesting

yukinarit (approved):
LGTM, thank you!

serde/core.py
acolley-gel: Yeah I wasn't aware of this myself.

Issue #418, opened 2023-07-25
Drop python 3.7 support


yukinarit:
As it's already EOL https://endoflife.date/python

PR #452, opened 2023-12-08
Fix `skip_if` causing serialization of a recursive type to be `O(2**n)` instead of `O(n)`

### LINKED ISSUE #451: Serialization of recursive data structures is accidentally exponential when using `skip_if`
hi!

passing `skip_if` causes serialization time for a recursive data structure to change from $\mathcal{O}(n)$ to $\mathcal{O}(2^n)$, where $n$ is the height of the recursive data structure.

minimal example:

```python
from __future__ import annotations

import time
from functools import reduce

import serde


@serde.serde
@dataclass
class Leaf:
    foo: str


@dataclass
class Tree:
    children: list[Tree | Leaf] = serde.field(skip_if=lambda _: False)


serde.serde(Tree)


def benchmark(height: int) -> float:
    tree = reduce(
        lambda child, _: Tree([child]),
        range(height - 1),
        Tree([]),
    )
    t0 = time.perf_counter()
    serde.to_dict(tree)
    return time.perf_counter() - t0


if __name__ == "__main__":
    print("height += 1 causes serialization time to double! accidentally exponential!:")
    for height in range(10, 20):
        dt = benchmark(height)
        print(f"{height = } -> {dt = :.3e}")
```

that is, there is some processing for which the root node gets processed once, its children get processed twice each, their children get processed four times each, and so forth. accidentally exponential!

gschaffner:
fixes #451.

i was not sure where to add a test for this, as it's an (exponential) performance issue rather than a correctness issue. a pointer would be much appreciated!

note that once Python 3.7 support is dropped (#418), this is a perfect use-case for the `:=` operator.

### DIFF
--- a/serde/se.py
+++ b/serde/se.py
@@ -553,8 +553,9 @@ def {{func}}(obj, reuse_instances = None, convert_sets = None):
   {% for f in fields -%}
   {% if not f.skip -%}
     {% if f.skip_if -%}
-  if not {{f.skip_if.name}}({{f|rvalue}}):
-    {{f|lvalue}} = {{f|rvalue}}
+  subres = {{f|rvalue}}
+  if not {{f.skip_if.name}}(subres):
+    {{f|lvalue}} = subres
     {% else -%}
   {{f|lvalue}} = {{f|rvalue}}
     {% endif -%}


yukinarit (approved):
LGTM!

PR #482, opened 2024-02-29
skip conversion if the object is already the correct type

uyha:
Motivation:
The `msgpack` library has a time called `Timestamp` which is an extension type from the msgpack spec. The library automatically detects it and deserializes the data to `Timestamp`. Hence, the conversion by `pyserde` is not needed.


yukinarit (approved):
Thank you!

PR #485, opened 2024-03-04
prefer `is_bearable` to `isinstance`

rtbs-dev:
Ran into an issue with parametrized generics not being supported by `isinstance`. Since you're already relying on beartype, I've found it _much_ easier to strictly rely on `is_bearable` to do dispatch, rather than `isinstance`. Stuff just behaves like I expect it to. 

Line 389 in `se.py` was raising a `SerdeError: isinstance() argument 2 cannot be a parametrized generic`. 

For more on the DOOR API see [here](https://beartype.readthedocs.io/en/latest/api_door/#door-procedures), with `is_bearable` specifically documented [here](https://beartype.readthedocs.io/en/latest/api_door/#beartype.door.is_bearable) As it mentions there, this is designed to replace `isinstance` by being a strict superset: 

> [is_bearable()](https://beartype.readthedocs.io/en/latest/api_door/#beartype.door.is_bearable) is a strict superset of the [isinstance()](https://docs.python.org/3/library/functions.html#isinstance) builtin. [is_bearable()](https://beartype.readthedocs.io/en/latest/api_door/#beartype.door.is_bearable) can thus be safely called wherever [isinstance()](https://docs.python.org/3/library/functions.html#isinstance) is called with the same exact parameters in the same exact order:

### DIFF
--- a/serde/se.py
+++ b/serde/se.py
@@ -21,6 +21,7 @@ from beartype.typing import (
     Type,
     Union,
 )
+from beartype.door import is_bearable
 
 import jinja2
 from typing_extensions import dataclass_transform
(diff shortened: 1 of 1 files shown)


yukinarit (approved):
LGTM

Issue #524, opened 2024-05-09
Performance becoming an issue


AustinScola:
Hi @yukinarit,

I've been loving using pyserde the last 2ish years, however recently I noticed that the import time for a lot of module's in a repo is quite slow. Using python's `-X importtime` I narrowed it down to modules that have a lot of pyserde decorated dataclasses. We are up to between 5-10seconds of import time being due to `pyserde` :(

Here is an example of how long some very simple dataclasses take to decorate:
```
from dataclasses import dataclass
import time

from serde import serde

@dataclass
class Foo:
    bar: int

start = time.time()
serde(Foo)
end = time.time()
print(end - start) # 0.019263029098510742


@dataclass
class Wobble:
    wham: int

@dataclass
class Wibble:
    wobble: Wobble

start = time.time()
serde(Wibble)
end = time.time()
print(end - start) # 0.033370018005371094
```

yukinarit:
That is a problem. Let me take a look.

yukinarit:
Looks like jinja2 is slow

```
Sat May 11 21:24:41 2024    profile_results.prof

         50295 function calls (47272 primitive calls) in 0.031 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2863    0.003    0.000    0.005    0.000 /home/yukinari/.cache/pypoetry/virtualenvs/pyserde-3viUZWFG-py3.10/lib/python3.10/site-packages/jinja2/nodes.py:167(iter_child_nodes)
        5    0.002    0.000    0.002    0.000 {built-in method builtins.compile}
      400    0.002    0.000    0.003    0.000 /home/yukinari/.cache/pypoetry/virtualenvs/pyserde-3viUZWFG-py3.10/lib/python3.10/site-packages/jinja2/lexer.py:667(tokeniter)
    10803    0.002    0.000    0.002    0.000 {built-in method builtins.isinstance}
     5167    0.002    0.000    0.002    0.000 /home/yukinari/.cache/pypoetry/virtualenvs/pyserde-3viUZWFG-py3.10/lib/python3.10/site-packages/jinja2/nodes.py:145(iter_fields)
    860/5    0.001    0.000    0.013    0.003 /home/yukinari/.cache/pypoetry/virtualenvs/pyserde-3viUZWFG-py3.10/lib/python3.10/site-packages/jinja2/visitor.py:35(visit)
[...]

PR #528, opened 2024-05-26
Cache code templates to speed up serde codegen

yukinarit:
Usage
$ python profile_codegen.py

Before
13140368 function calls (12815954 primitive calls) in 5.964 seconds

After
6391286 function calls (6082792 primitive calls) in 3.396 seconds

#524


yukinarit:
Awesome! I will finish up the PR and hopefully merge within this week.

Issue #575, opened 2024-07-22
serde class decorator breaks forward reference resolution


jennydaman:
```python
import dataclasses
import typing

from serde import deserialize


@dataclasses.dataclass
class Expected:
    def fun(self) -> "SomeReturnClass":
        ...


@deserialize
class Actual:
    def fun(self) -> "SomeReturnClass":
        ...


class SomeReturnClass:
    pass


assert typing.get_type_hints(Expected.fun)['return'] is SomeReturnClass  # ok
assert typing.get_type_hints(Actual.fun)['return'] is SomeReturnClass    # fail
```

Test in Python versions 3.11.9 and 3.12.4, pyserde==0.19.3

jennydaman:
This has to do with beartype. For the purposes of my own code, I found a different workaround. I wanted to do type-narrowing on the type returned by `typing.get_type_hints(Actual.fun)['return']`, which was possible with `dataclasses.is_dataclass` and `dataclasses.fields`.

jennydaman:
Thank you for the tip, @yukinarit 

I did try that earlier, but even with `type_check=disabled` the forward reference is still mangled by beartype.

```python
import dataclasses
import typing

from serde import deserialize, disabled


@dataclasses.dataclass
class Expected:
    def fun(self) -> "SomeReturnClass":
        ...


@deserialize(type_check=disabled)
class Actual:
    def fun(self) -> "SomeReturnClass":
        ...


class SomeReturnClass:
    pass


def test_serde575():
    assert typing.get_type_hints(Expected.fun)['return'] is SomeReturnClass
    assert typing.get_type_hints(Actual.fun)['return'] is SomeReturnClass
```

Pytest output:

```
ssertionError: assert <forwardref SomeReturnClass(__name_beartype__='SomeReturnClass', __scope_name_beartype__='test_serde575')> is SomeReturnClass
```

jennydaman:
A workaround for beartype v0.19.0 is discussed here: https://github.com/beartype/beartype/issues/410#issuecomment-2249195428

@yukinarit the unexpected behavior remains but it's no longer a problem to me. Do you want to close the issue?

yukinarit:
@jennydaman 
if I add `type_check=disabled` to `Expected` class, the passes successfully

jennydaman:
It does not. Here is a fully reproducible copy-paste MRE

```shell
docker run --rm -i docker.io/library/python:3.12.4-bookworm bash << EOF
[...]

PR #577, opened 2024-07-27
type_check=disable should not implement beartype

yukinarit:
The issue was relevant only when `@serialize` or `@deserialize` used.

for #575

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -274,7 +274,7 @@ def deserialize(
         for typ in iter_types(cls):
             # When we encounter a dataclass not marked with deserialize, then also generate
             # deserialize functions for it.
-            if is_dataclass_without_de(typ):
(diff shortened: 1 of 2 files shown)


jennydaman (approved):

serde/de.py
```diff
@@ -274,7 +274,7 @@ def wrap(cls: type[T]) -> type[T]:
         for typ in iter_types(cls):
             # When we encounter a dataclass not marked with deserialize, then also generate
             # deserialize functions for it.
-            if is_dataclass_without_de(typ):
+            if is_dataclass_without_de(typ) and typ is not cls:
```
yukinarit, line 277: what was happening is, `@deserialize` was called twice. The second call here does not pass `type_check` in, so it became `type_check=strict`

I wonder we should pass all the class attributes from parent class :thinking:

PR #646, opened 2025-04-26
Fix skip_none parameter in several serialization modules

### LINKED ISSUE #645: `skip_none` is a no-op in JSON, YAML, and Pickle
`serde.json.to_json`, `serde.yaml.to_yaml`, and `serde.pickle.to_pickle` are all documented as – and accept – a `skip_none` argument. However, it's a no-op, as it's dropped on the floor instead of being passed to `as_dict`. It seems like only TOML fully implemented it. (And interestingly, MsgPack isn't documented as accepting that option.)

Compare [YAML](https://github.com/yukinarit/pyserde/blob/1e3b0698e81b6498f5360e4c56cc8d14cc477a74/serde/yaml.py#L51) with [TOML](https://github.com/yukinarit/pyserde/blob/1e3b0698e81b6498f5360e4c56cc8d14cc477a74/serde/toml.py#L64).

I don't mind to try and contribute a fix, but wanted to be sure it would be welcome before I spend the time. Thanks!

vtbassmatt:
Fixes #645 by passing `skip_none` to YAML/JSON serializers. The changes are demonstrated by new tests that fail without my change and pass with my change.

~~For Pickle, the argument didn't seem to make a difference, so I marked it as deprecated. No tests are introduced.~~ *Edit: this change adds multiple lines of "untested code" according to Codecov, and introducing tests just to prove it raises a warning seems pointless. I dropped this change.*

I also noticed that one of the TOML tests wasn't being called because its name didn't start with `test_`. And the second half of the test case wasn't asserting anything. I fixed that test module to match with the ones I introduced for YAML/JSON.

### DIFF
--- a/serde/json.py
+++ b/serde/json.py
@@ -76,7 +76,14 @@ def to_json(
     your own logic.
     """
     return se.serialize(
-        to_dict(obj, c=cls, reuse_instances=reuse_instances, convert_sets=convert_sets), **opts
+        to_dict(
+            obj,
+            c=cls,
+            reuse_instances=reuse_instances,
+            convert_sets=convert_sets,
+            skip_none=skip_none,
+        ),
+        **opts,
     )
 
 
--- a/serde/yaml.py
+++ b/serde/yaml.py
@@ -48,7 +48,14 @@ def to_yaml(
     your own logic.
     """
     return se.serialize(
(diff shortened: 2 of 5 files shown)


yukinarit (approved):
Thank you!

PR #649, opened 2025-04-29
Support `numpy>2.0.0`

### LINKED ISSUE #562: Support numpy v2

krylowicz:
This adds support for `numpy>2.0.0` by:

- Update the upper version restriction from `<2.0.0` to `<3.0.0` for all supported python versions
- Add [NPY201](https://docs.astral.sh/ruff/rules/numpy2-deprecation/) ruff rule for reporting uses of functions and constants deprecated in NumPy 2.0
- Remove usage of `np.float_` (which is a direct [alias](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#main-namespace:~:text=get_array_wrap-,float_,-Use%20np.float64) for `np.float64`) from tests and numpy example 

Closes #562

### DIFF
--- a/examples/type_numpy.py
+++ b/examples/type_numpy.py
@@ -11,7 +11,7 @@ class Foo:
     in_: numpy.int_
     inc: numpy.intc
     ui: numpy.uint
-    fl: numpy.float_
+    fl: numpy.float64
     st: numpy.str_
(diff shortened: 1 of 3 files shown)


yukinarit (approved):
LGTM! Thanks!

Issue #652, opened 2025-08-07
CBOR support


davetapley:
Would you accept a PR to add [CBOR](https://cborbook.com/introduction/cbor_vs_the_other_guys.html) support?

[cbor2](https://github.com/agronholm/cbor2) claims
> Simple api like json or pickle modules.

So I'm going to see if I can just copy [`json.py`](https://github.com/yukinarit/pyserde/blob/22d0b40d5ecbfbc048bb94978ac9c528a713af38/serde/json.py). Hopefully not too complicated.

davetapley:
@yukinarit I couldn't find good enough performance gains to justify it in the end.

PR #653, opened 2025-08-08
CBOR support

davetapley:
- https://github.com/yukinarit/pyserde/issues/652

Draft for feedback. No tests yet, but it seems to work.


davetapley:
Close for separate package per https://github.com/yukinarit/pyserde/issues/652#issuecomment-3171894035

PR #660, opened 2025-08-22
Support customizing tagged union tag values

mickvangelderen:


serde/core.py
```diff
@@ -776,6 +777,13 @@ class Kind(enum.Enum):
     tag: Optional[str] = None
     content: Optional[str] = None
     kind: Kind = Kind.External
+    tags: Optional[frozendict[type[Any], str]] = None
```
mickvangelderen, line 780: Using frozendict so the value can be hashed.

serde/de.py
```diff
@@ -1129,18 +1129,18 @@ def {{func}}(cls=cls, maybe_generic=None, maybe_generic_type_vars=None, data=Non
   try:
     # create fake dict so we can reuse the normal render function
     {% if tagging.is_external() and is_taggable(t)  %}
-    ensure("{{typename(t)}}" in data , "'{{typename(t)}}' key is not present")
-    fake_dict = {"fake_key": data["{{typename(t)}}"]}
+    ensure("{{tagging.tag_for(t)}}" in data , "'{{tagging.tag_for(t)}}' key is not present")
+    fake_dict = {"fake_key": data["{{tagging.tag_for(t)}}"]}
```
mickvangelderen, line 1133: I wonder if we could inline the tags as literals, rather than looking them up during deserialization.

serde/core.py
```diff
@@ -825,39 +838,58 @@ def __call__(self, cls: T) -> _WithTagging[T]:
...
     else:
         return tagging
 
 
+# TODO: This is an instance rather than a function for backwards-compatibility
+# reasons. In the next major version increase this should be replaced with a
+# function.
 ExternalTagging = Tagging()
 
 Untagged = Tagging(kind=Tagging.Kind.Untagged)
```
mickvangelderen, line 895: This is unfortunate, not sure what to call the function.

PR #661, opened 2025-08-25
Disable fail-fast

mickvangelderen:
Allows you to see for all python versions if they succeed or fail at once, which is useful information.


yukinarit (approved):
Thanks @mickvangelderen and apologize for the delay. 
I will take a look at your PRs little by little.

PR #665, opened 2025-08-25
Cache through type rather than string

mickvangelderen:
First merge https://github.com/yukinarit/pyserde/pull/662

This implements caching through the type or type construct (with tagging information) rather than trying to serialize that information into a string. This fixes the issue illustrated by https://github.com/yukinarit/pyserde/pull/664


yukinarit (approved):
LGTM, thank you!

PR #739, opened 2026-03-31
skip ignored fields in compat walkers

AllanChain:
This is a follow-up for #722. The code is generated by LLM, but I have reviewed it.

This patch makes the recursive walkers in `serde.compat` respect `serde_skip` for:
- `iter_types()`
- `iter_literals()`

### 1. Recursive PEP 695 alias on a skipped field

On Python 3.12+, this fails during `@serde` decoration because `iter_types()` expands the skipped field type anyway and recurses forever:

```python
type NestedDict = dict[str, NestedDict | int]

@serde
class Config:
    name: str
    internal: NestedDict | None = field(default=None, skip=True)
```

Observed failure:

- RecursionError from serde.compat.iter_types()

### 2. Skipped `Literal[EnumMember]` field

This also should be harmless, but iter_literals() still walked the skipped field and generated a literal deserializer for it:

```python
class Choice(enum.Enum):
  X = "x"

@serde
class Foo:
    name: str
    ignored: Literal[Choice.X] = serde.field(default=Choice.X, skip=True)
```

Without this fix, pyserde can fail at decoration/codegen time because enum-valued literals render invalid Python in the generated literal helper.

## Fix

When recursively visiting dataclass fields in `serde.compat`, skip fields with `f.metadata["serde_skip"]`.

That keeps ignored fields out of the internal type/literal discovery phase, which matches the intended runtime behavior: ignored fields should not affect generated serde code.

## Tests

Added regressions for both user-visible failures:

- skipped recursive PEP 695 alias field
- skipped `Literal[EnumMember]` field

### DIFF
--- a/serde/compat.py
+++ b/serde/compat.py
@@ -353,7 +353,8 @@ def iter_types(cls: type[Any]) -> list[type[Any]]:
             lst.add(cls)
             if isinstance(cls, type):
                 for f in dataclass_fields(cls):
-                    recursive(f.type)
+                    if not f.metadata.get("serde_skip"):
+                        recursive(f.type)
         elif isinstance(cls, str):
             lst.add(cls)
         elif is_opt(cls):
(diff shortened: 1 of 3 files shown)


yukinarit (approved):
LGTM, Thank you!

PR #746, opened 2026-05-05
Reject None for non-optional field in coerce mode

AllanChain:
Currently, `type_check=coerce` lets the deserializer silently accept `None` for a non-optional field, which is inconsistent with the documentation. For example:

```python
import serde

@serde.serde(type_check=serde.coerce)
class A:
    a: int = 0

print(serde.from_dict(A, {"a": None}))
```

The above code outputs the following before the fix:

```python
A(a=None)
```

But a `SerdeError` is expected.

Here, we raise `SerdeError` when `coerce_object()` receives `None` for a non-optional field.


yukinarit (approved):
Thank you 🙏

PR #747, opened 2026-05-05
Improve missing field error messages

AllanChain:
Missing required fields currently surface as a bare `SerdeError("'field'")`. For example,

```python
import serde


@serde.serde()
class A:
    a: int


print(serde.from_dict(A, {}))
```

The above code raises the following error before the fix:

```
serde.compat.SerdeError: 'a'
```

Here we catch that `KeyError` at the required-field lookup site, so pyserde raises a contextual `SerdeError` with the field and class name while leaving defaulted-field fallback behavior unchanged.

After the fix, the message becomes:

```
serde.compat.SerdeError: missing required field 'a' while deserializing A
```

---

The fix was done with the assistance of LLM, and I reviewed and tested it. It looks reasonable, but I'm not certain if the fix is ideal.

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -1288,10 +1288,18 @@ def {{func}}(cls=cls, maybe_generic=None, maybe_generic_type_vars=None, data=Non
   {% endif %}
 
   {% for f in fields %}
+  {% set field_arg = arg(f,loop.index-1) %}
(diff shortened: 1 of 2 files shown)


yukinarit (approved):
Nice, thanks!

PR #751, opened 2026-05-28
Fix deserialization with `skip`

PingPongun:
`pyserde` currently fails to deserialize when after field with `skip` attribute there is a non skipped field. (ofc in simple cases, skipped fields could be reordered to be the last ones, but this is not a solution when inheritance is in game, also I've not noticed such limitation of `skip` being noted anywhere in documentation)

This PR adds such case as test scenario ( [test failed before fix](https://github.com/PingPongun/pyserde/actions/runs/26597183001/job/78371134687#step:6:10573) ) and fixes the problem, by forcing keyword based class initialization (for fields after skipped field)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: Tests
 
-on: [pull_request]
+on: [pull_request, workflow_dispatch]
 
 jobs:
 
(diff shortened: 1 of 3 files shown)


yukinarit (approved):
Thanks!

PR #752, opened 2026-05-31
Fix flattened child unknown fields

AllanChain:
Previously, when we have a flattened child class with `deny_unknown_fields=True`, we will have a problem deserializing.

```python
import serde

@serde.serde(deny_unknown_fields=True)
class A:
    a: int

@serde.serde
class B:
    b: int
    a: A = serde.field(flatten=True, default_factory=A)

serde.from_dict(B, {"a": 1, "b": 2})
```

Actual behavior:

```text
serde.compat.SerdeError: unknown fields: {'b'}, expected one of {'a'}
```

Expected:

```python
B(b=2, a=A(a=1))
```

The fix strategy is passing only the relevant flattened input to child deserializers, achieved by knowing the exact keys expected by the flattened child. Special cases are flattened dict fields. If the parent has flattened dict fields whereas the flattened child does not, we will keep the unknown fields only visible by the parent. Otherwise, we will pass those to the child.

---

The changes are done by a coding agent, and I have reviewed the code by trying to understand the changes line by line. The changes are reasonable and readable.

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -258,6 +258,12 @@ class Scope:
     defaults: dict[str, Callable[..., Any] | Any] = dataclasses.field(default_factory=dict)
     """ Default values of the dataclass fields (factories & normal values) """
 
+    de_known_fields: set[str] = dataclasses.field(default_factory=set)
+    """ Field names accepted by the generated from_dict function """
+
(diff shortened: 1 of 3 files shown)


yukinarit (approved):
LGTM, thanks!

PR #754, opened 2026-06-07
Fix serialization of PEP 695 type-alias unions

### LINKED ISSUE #753: Serialization with union type defined by type-statement is broken
The type statement has been supported since #669.
However, when serialization with a union type defined by a type statement, it will fail to recognize it as a union.
Note that, for the type statement `type T = A | B`, it creates a `T` with `type(T) == TypeAliasType`.
I think this needs special care in the serialization path.

Reproduce code:
``` python
from typing import TypeAlias

import serde.json
import serde


@serde.serde
class Foo:
    a: int = 123


@serde.serde
class Dummy:
    msg = "hello"


type T1 = Foo | Dummy
T2: TypeAlias = Foo | Dummy

a = Foo()

print("no-cls")
print(serde.json.to_json(a))

print(f"cls=T1, {type(T1)=}")
print(serde.json.to_json(a, cls=T1))

print(f"cls=T2, {type(T2)=}")
print(serde.json.to_json(a, cls=T2))

print(f"cls=Lit, {type(Foo | Dummy)=}")
print(serde.json.to_json(a, cls=Foo | Dummy))

```

Python version and output:
```
$ python --version; python test.py
Python 3.14.4
no-cls
{"a":123}
cls=T1, type(T1)=<class 'typing.TypeAliasType'>
{"a":123}
cls=T2, type(T2)=<class 'typing.Union'>
{"Foo":{"a":123}}
cls=Lit, type(Foo | Dummy)=<class 'typing.Union'>
{"Foo":{"a":123}}

```

gaoflow:
Fixes #753.

A PEP 695 `type T = Foo | Bar` alias is a `TypeAliasType`, not the union it aliases. When such an alias was passed as the top-level `cls`, `to_obj` / `from_obj` failed the `is_union(c)` check and fell through to plain (untagged) serialization, diverging from `cls=Foo | Bar`:

```python
type T = Foo | Dummy

to_json(Foo(), cls=T)          # {"a":123}            <- wrong, no tagging
to_json(Foo(), cls=Foo | Dummy) # {"Foo":{"a":123}}   <- expected
```

### Fix
Unwrap the alias to its `__value__` at the top of `to_obj` (se.py) and `from_obj` (de.py), so it is recognized as the union it stands for. This mirrors how PEP 695 aliases are already unwrapped on the field-rendering path (added in #669). Now `cls=T` matches `cls=Foo | Dummy` for both serialization and deserialization round-trips.

### Tests
Added `test_pep695_type_alias_as_top_level_union` in `tests/test_type_alias.py` (fails without the patch). Full suite green (5169 passed, excluding the numpy/jaxtyping-gated files).

Thanks to @Luffbee for the clear report and reproducer.

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -456,6 +456,11 @@ def from_obj(
 
     res: Any
 
+    # A PEP 695 type alias (``type T = Foo | Bar``) is a TypeAliasType, not the
+    # aliased type itself, so unwrap it to recognize e.g. a union.
+    if is_pep695_type_alias(c):
+        c = c.__value__  # type: ignore[attr-defined]
+
     # It is possible that the parser already generates the correct data type requested
     # by the caller. Hence, it should be checked early to avoid doing anymore work.
     if type(o) is c:
--- a/serde/se.py
+++ b/serde/se.py
@@ -402,6 +402,11 @@ def to_obj(
             skip_none=skip_none,
         )
 
+        # A PEP 695 type alias (``type T = Foo | Bar``) is a TypeAliasType, not
+        # the aliased type itself, so unwrap it to recognize e.g. a union.
+        if c is not None and is_pep695_type_alias(c):
+            c = c.__value__
+
         # If a class in the argument is a non-dataclass class e.g. Union[Foo, Bar],
         # pyserde generates a wrapper (de)serializable dataclass on the fly,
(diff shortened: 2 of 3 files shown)


yukinarit (approved):
LGTM!

PR #756, opened 2026-06-07
Fix custom field serializer inside a flattened field

### LINKED ISSUE #453: Custom field serialzer in flattened field bug
It appears that a custom field serializer inside of a flattened field fails to serialzie:

Example:
```python3
from serde import field, serialize, to_dict

@serialize
class Child:
    value: int = field(serializer=str)

@serialize
class Parent:
    child: Child = field(flatten=True)

parent = Parent(child=Child(value=3))
print(to_dict(parent))
```

Stacktrace:
```
  File "/private/tmp/foo.py", line 12, in <module>
    print(to_dict(parent))
          ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/serde/se.py", line 450, in to_dict
    return to_obj(  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/serde/se.py", line 378, in to_obj
    raise SerdeError(e) from None
serde.compat.SerdeError: name 'value_serializer' is not defined
```

gaoflow:
Fixes #453.

A custom field serializer on a field of a flattened dataclass crashed serialization:

```python
@serialize
class Child:
    value: int = field(serializer=str)

@serialize
class Parent:
    child: Child = field(flatten=True)

to_dict(Parent(child=Child(value=3)))  # SerdeError: name 'value_serializer' is not defined
```

### Cause
A flattened dataclass field is rendered *inline* in the enclosing class's generated serialize function, so it emits references like `value_serializer(...)`. But the loop that registers custom field serializers (and `skip_if` predicates) in the generated globals only iterated the top-level fields — the flattened child's serializer was never put in scope.

### Fix
Recurse into flattened (and optional-flattened) dataclass fields when collecting field serializers / `skip_if` predicates, so they are available in the enclosing scope. Handles nested flattens too.

### Tests
Added `test_flatten_custom_field_serializer` (basic + nested flatten; fails without the patch). Full suite green (5169 passed). Deserialization was already unaffected (a flattened child is deserialized via its own scoped function).

Thanks @AustinScola for the report and minimal reproducer.

### DIFF
--- a/serde/se.py
+++ b/serde/se.py
@@ -307,11 +307,7 @@ def serialize(
             add_func(scope, union_key, render_union_func(cls, union_args, tagging), g)
             scope.union_se_args[union_key] = union_args
 
-        for f in sefields(cls, serialize_class_var):
-            if f.skip_if:
-                g[f.skip_if.name] = f.skip_if
-            if f.serializer:
-                g[f.serializer.name] = f.serializer
+        add_field_serializers_to_scope(sefields(cls, serialize_class_var), g, serialize_class_var)
 
         add_func(
             scope,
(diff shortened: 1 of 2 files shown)


yukinarit (approved):
LGTM, thanks!

PR #762, opened 2026-06-08
Apply global class (de)serializers to top-level values (fixes #514)

### LINKED ISSUE #514: Custom Serializer doesn't work with top-level data.

gaoflow:
Fixes #514.

### Problem

A custom serializer registered globally with `serde.add_serializer` (and the matching `add_deserializer`) is only consulted when the custom type appears as a **dataclass field**. When the same type is passed **directly** to `to_dict`/`to_tuple`/`from_dict`, the registration is ignored:

```python
serde.add_serializer(MySerializer())      # serialize(self, value: MyClass) -> dict
serde.add_deserializer(MyDeserializer())

serde.to_dict(MyWrapper(MyClass(1, 2)))   # {'value': {'a': 1, 'b': 2}}  ✓ nested works
serde.to_dict(MyClass(1, 2))              # <MyClass object ...>          ✗ returns the raw object
```

`to_dict(MyClass(...))` returns the object unchanged instead of `{'a': 1, 'b': 2}`, and `from_dict(MyClass, {...})` raises `TypeError` while trying to construct the class directly.

### Root cause

`to_obj` (se.py) and `from_obj` (de.py) handle dataclasses, containers, unions, str/datetime, etc., but neither checks the global `ClassSerializer`/`ClassDeserializer` registry for the top-level value's type. The generated field-rendering code (`Renderer.render` / `Defactory.render`) *does* check it (and gives it precedence), so nested fields work while the top level falls through to `return o` / `c(o)`.

### Fix

Add `find_global_class_serializer(type)` / `find_global_class_deserializer(type)` helpers that reuse the exact same plum-dispatch lookup the field renderers use (`serialize.methods` / `deserialize.methods` → value type). `to_obj`/`from_obj` consult them for the value's type, matching the nested precedence so behaviour is identical whether the custom type is nested or top-level. When no global handler matches the type, the registry is empty and the lookup is a no-op, so existing behaviour is unchanged.

### Verification

- New `test_global_class_serializer_top_level` in `tests/test_custom.py`: registers a global serializer + deserializer, asserts top-level `to_dict`/`to_tuple`/`from_dict` round-trip and that the nested case still matches (snapshots/restores the global registry so it doesn't leak into other tests). Red on `main`, green with the fix.
- Full suite green (`pytest tests`, excluding the jaxtyping-gated numpy/code-completion files): 5174 passed.
- `pre-commit` (black, ruff, pyright, mypy) clean on all changed files.

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -464,6 +464,22 @@ class Deserializer(Generic[T], metaclass=abc.ABCMeta):
         raise NotImplementedError
 
 
+def find_global_class_deserializer(typ: type[Any]) -> ClassDeserializer | None:
+    """Return a globally registered class deserializer that handles ``typ``, if any.
+
+    This mirrors the lookup used when rendering dataclass fields, so a deserializer
+    registered with ``serde.add_deserializer`` also applies when ``typ`` is passed
+    directly to ``from_dict``/``from_tuple`` instead of nested in a dataclass
+    (https://github.com/yukinarit/pyserde/issues/514).
+    """
+    for class_deserializer in GLOBAL_CLASS_DESERIALIZER:
+        for method in class_deserializer.__class__.deserialize.methods:  # type: ignore[attr-defined]
+            # signature.types[1] is ``type[X]``; unwrap it to the value type ``X``.
+            if get_args(method.signature.types[1])[0] is typ:
+                return class_deserializer
+    return None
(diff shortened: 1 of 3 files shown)


yukinarit (approved):
Thanks, LGTM!

PR #763, opened 2026-06-12
Fix JSON deserialization of IntEnum/IntFlag dict keys

sarathfrancis90:
Round-tripping a dataclass with a `dict` keyed by an `IntEnum` (or `IntFlag`) through JSON fails:

```python
@serde
@dataclass
class C:
    d: dict[Num, str]  # Num is an IntEnum

obj = C({Num.ONE: "a"})
from_json(C, to_json(obj))
# SerdeError: failed to coerce the field C.k value 1 into Num: '1' is not a valid Num
```

`to_json` works fine, and the dict round-trip (`from_dict`/`to_dict`) works. The problem is JSON-only: JSON forces object keys to strings, so the key comes back as `"1"`, and the generated code calls `Num("1")`, which raises even though the value is valid. `dict[Color, str]` with a str-valued `Enum` already round-trips, so int-valued enum keys were the odd one out.

I route enum deserialization through a small `deserialize_enum` helper that, when the direct lookup fails for a `str`, retries after coercing the value into the enum's member value type. str enums and value-position enums are untouched; invalid values still raise `SerdeError`. Added a parametrized test alongside `test_dict_with_non_str_keys` (msgpack/toml skipped since they can't represent the keys). Full suite green.

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -1157,12 +1157,13 @@ def deserialize_enum(typ: type[enum.Enum], value: Any) -> enum.Enum:
     try:
         return typ(value)
     except (ValueError, KeyError):
+        # A ValueError/KeyError here means ``typ`` has at least one member (an
+        # empty enum would raise TypeError instead), so ``__members__`` is never
+        # empty in this branch.
         if isinstance(value, str):
-            members = list(typ.__members__.values())
-            if members:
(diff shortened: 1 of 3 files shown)


yukinarit (approved):
LGTM

PR #764, opened 2026-06-17
Fix Union[None, X] being serialized as null after Optional[X]

gaoflow:
### Summary

A field typed `Union[None, X]` is silently serialized as `null` whenever any `Optional[X]` field was processed earlier in the program:

```python
@serde
@dataclass
class HasOptional:
    v: Optional[int]

@serde
@dataclass
class NoneFirst:
    v: Union[None, int]

to_dict(NoneFirst(7))   # -> {'v': None}   (expected {'v': 7})
to_json(NoneFirst(7))   # -> '{"v":null}'  (expected '{"v":7}')
```

`Union[None, int]` is the same type as `Optional[int]`, just with its arguments in the opposite order, so collapsing every value to `null` is silent data loss. Decorating `NoneFirst` on its own (without an `Optional` field anywhere before it) produces the correct result, which is what makes this order-dependent and easy to miss.

### Root cause

`compat.is_opt()` classifies a type as optional by inspecting argument *positions*:

```python
... and not is_none(args[0]) and is_none(args[1])
```

so its result depends on the union argument order. It was decorated with `@cache`, but `typing.Union` hashes and compares equal regardless of argument order, so `Optional[int]` (`== Union[int, None]`) and `Union[None, int]` share one cache entry. The first call wins: once an `Optional` field caches `True`, every later `Union[None, int]` gets that cached `True`, is treated as an optional whose inner type is its first argument (`NoneType`), and serializes to `null`.

### Fix

Drop the `@cache` on `is_opt()` so each ordering is evaluated on its own. The function only does a couple of cheap checks and is used during type analysis, not in a per-record hot path, so this has no meaningful performance impact. Added `test_union_none_first_argument_order`, which fails on `main` and passes with the fix; the full suite (`5228 passed, 1 skipped`) is unaffected.

I used an AI assistant to help investigate and implement this under my direction.

### DIFF
--- a/serde/compat.py
+++ b/serde/compat.py
@@ -521,7 +521,10 @@ def is_union(typ: Any) -> bool:
     return typing_inspect.is_union_type(typ)  # type: ignore
 
 
-@cache
+# Not memoized: the result depends on the order of the union arguments
+# (``args[0]`` must be non-None and ``args[1]`` must be None), but ``Union``
+# hashes and compares equal regardless of that order, so a cache would return
+# the first-seen answer for both ``Optional[int]`` and ``Union[None, int]``.
 def is_opt(typ: Any) -> bool:
     """
     Test if the type is `typing.Optional`.
--- a/serde/compat.py
+++ b/serde/compat.py
@@ -574,7 +574,8 @@ def is_bare_opt(typ: Any) -> bool:
     return not type_args(typ) and typ is Optional
 
 
-@cache
+# Not memoized: like ``is_opt``, the result depends on the order of union
+# arguments even though ``Optional[T]`` and ``Union[None, T]`` compare equal.
 def is_opt_dataclass(typ: Any) -> bool:
     """
     Test if the type is optional dataclass.
@@ -587,7 +588,7 @@ def is_opt_dataclass(typ: Any) -> bool:
     >>> is_opt_dataclass(Foo)
     False
     >>> is_opt_dataclass(Optional[Foo])
-    False
+    True
     """
     args = get_args(typ)
     return is_opt(typ) and len(args) > 0 and is_dataclass(args[0])
--- a/serde/de.py
+++ b/serde/de.py
@@ -482,6 +482,16 @@ def find_global_class_deserializer(typ: type[Any]) -> ClassDeserializer | None:
     return None
 
 
+def _deserialize_optional(
+    c: type[Any],
+    o: Any,
+    deserializer: Callable[[type[Any], Any], Any],
+) -> Any:
+    if o is None:
+        return None
+    return deserializer(type_args(c)[0], o)
+
+
 def from_obj(
     c: type[T],
     o: Any,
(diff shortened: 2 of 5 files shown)


binggao1230:
Heads-up on the two red checks: both look pre-existing rather than caused by this change (which only touches `serde/compat.py`).

- **Check formatting**: ruff reports `C901 \`from_obj\` is too complex (31 > 30)` in `serde/de.py`. This reproduces on a clean `main` checkout, so it is independent of this PR.
- **Python 3.14**: the failing test is `test_global_class_serializer_top_level`, which only trips under parallel `-n auto`. It passes in isolation and in a local full `pytest -n auto` run on 3.14 here, so it looks like a test-ordering sensitivity around the global serializer registry rather than a regression from this change.

I can open a small separate PR to bring `from_obj` back under the complexity threshold if that would be helpful.

binggao1230:
Update: I pushed two small follow-up commits to keep this PR green on the current CI baseline:

- reduced `from_obj`'s ruff complexity below the configured threshold
- adjusted the global class serializer tests to avoid `plum`'s intermittent Python 3.14.6/Linux handling of generic `dict[str, Any]` dispatch annotations; the tests still cover the pyserde behavior under test, which is that global class (de)serializers are used for top-level values

Local verification included:

```bash
uv run ruff check serde/de.py tests/test_custom.py
uv run ruff format --check serde/de.py tests/test_custom.py
uv run pyright tests/test_custom.py
uv sync --python 3.14 --extra sqlalchemy --frozen
uv run --python 3.14 pytest tests --doctest-modules serde -q -nauto
```

The latest GitHub Actions run is now green across Python 3.10-3.14, formatting, coverage, and Codecov.

serde/compat.py
```diff
@@ -521,7 +521,10 @@ def is_union(typ: Any) -> bool:
     return typing_inspect.is_union_type(typ)  # type: ignore
 
 
-@cache
+# Not memoized: the result depends on the order of the union arguments
+# (``args[0]`` must be non-None and ``args[1]`` must be None), but ``Union``
+# hashes and compares equal regardless of that order, so a cache would return
+# the first-seen answer for both ``Optional[int]`` and ``Union[None, int]``.
```
yukinarit, line 527: is_opt_dataclass also needs to be fixed? WDYT?

binggao1230: Yes, agreed. `is_opt_dataclass` had the same order-sensitive cache issue because `Optional[Foo]` and `Union[None, Foo]` compare equal and share a hash, while `is_opt(...)` intentionally depends on the original union argument order.

I pushed `9d2150b` to remove that cache as well and added a regression test that calls `is_opt_dataclass(Optional[Foo])` before `is_opt_dataclass(Union[None, Foo])`, then repeats the reverse-sensitive assertions.

Verified with:
- `python -m pytest tests/test_compat.py::test_is_opt_dataclass_preserves_union_arg_order tests/test_union.py::test_union_none_first_argument_order -q`
- `python -m doctest serde/compat.py`
- `ruff check serde/compat.py tests/test_compat.py`

binggao1230:
FYI on the remaining red `claude-review` check: it failed before doing a code review because the Claude action could not fetch an OIDC token:

```text
Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
Could not fetch an OIDC token. Did you remember to add `id-token: write` to your workflow permissions?
```

The Python matrix, formatting, coverage, and Codecov checks are green on `9d2150b`.

serde/compat.py
yukinarit: thanks

yukinarit (approved):
LGTM, thanks!

Issue #585, opened 2024-08-19
Enum deserialization doesn't work with non-simple enum values


morrison12:
`Enum`s with "complex" values, for example a `tuple`, can't be deserialized as shown from the example below.
In the example below, it appears it is because the values are deserialized as a` list` but only a `tuple` is permitted !?!
If the name and not the value(s) were serialized for `Enum`s this could be avoided but would break the cases where the value is desired.  Perhaps serialize the name for all but `IntEnum` and `StrEnum` ?

# Example
```python
"""
Example of serializing and deserializing an Enum with a "complex" value
"""

from dataclasses import dataclass
from enum import Enum, unique

from serde import serde
from serde.json import from_json, to_json


@unique
class B(Enum):
    X = ("one", "some info on one")
    Y = ("two", "some info on two")
    Z = ("three", "some info on three")


@serde
@dataclass
class C:
    m: B


original = C(B.Y)

as_json = to_json(original)
print(f"{as_json=}")
print(f"{B.Z.value=}")
nb = B(("three", "some info on three"))
print(f"{nb.value=}")
from_json = from_json(C, as_json)
print("de-serialized as:", as_json)

assert original == from_json
```

# Output
```
as_json='{"m":["two","some info on two"]}'
B.Z.value=('three', 'some info on three')
nb.value=('three', 'some info on three')
Traceback (most recent call last):
  File "/Users/james/Projects/spt/examples/ts9.py", line 32, in <module>
    from_json = from_json(C, as_json)
                ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/james/.virtualenvs/spt/lib/python3.11/site-packages/serde/json.py", line 105, in from_json
    return from_dict(c, de.deserialize(s, **opts), reuse_instances=False)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/james/.virtualenvs/spt/lib/python3.11/site-packages/serde/de.py", line 533, in from_dict
    return from_obj(cls, o, named=True, reuse_instances=reuse_instances)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/james/.virtualenvs/spt/lib/python3.11/site-packages/serde/de.py", line 501, in from_obj
    raise SerdeError(e) from None
[...]

PR #765, opened 2026-06-17
Fix enum with tuple values deserialization

luxShrine:
First noted in (#585), when a tuple field of an enum was serialized into a list, it would not be deserialized into a tuple. This fix aims to patch that gap, by making a minimal change to the enum deserializing helper function, simply checking if the enum field is a tuple and converting it from a list if so. 

This is a small patch, and in the original discussion it was noted that a more robust system would be to have a setup that would generate "(de)serialize functions for enum", not unlike what is done currently for `Union`. This patch for example only covers a single depth of tuples: `("a", ("b", "c"))` serializes as ` ["a", ["b", "c"]]` and deserializes as `("a", ["b", "c"])` which would fail to properly align with the original enum value. However, the value in making a small fix like this seemed worthwhile to patch, and leaves the case of more complex fields to the more substantial project of generating deserialization functions for enums.

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -1160,10 +1160,11 @@ def deserialize_enum(typ: type[enum.Enum], value: Any) -> enum.Enum:
         # A ValueError/KeyError here means ``typ`` has at least one member (an
         # empty enum would raise TypeError instead), so ``__members__`` is never
         # empty in this branch.
-        if isinstance(value, str):
-            member_value_type = type(next(iter(typ.__members__.values())).value)
-            if member_value_type is not str:
-                return typ(member_value_type(value))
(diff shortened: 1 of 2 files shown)


binggao1230:
The failing `Check formatting` job is Ruff `C901` on `serde/de.py::from_obj`, which is unchanged in this PR. I reproduced the same failure on current `main` with:

```bash
UV_FROZEN=1 uv run --frozen ruff check --force-exclude serde/de.py
```

I opened #767 to fix that main-branch lint regression separately. That PR is now green, so this PR should not need a business-logic change for the formatting failure.

yukinarit (approved):
LGTM, thanks! 🙏

PR #766, opened 2026-06-19
Fix JSON deserialization of int/float dict keys

sarathfrancis90:
A dataclass with a `dict[int, V]` or `dict[float, V]` field serializes to JSON fine, but can't be deserialized back:

```python
@serde
@dataclass
class Foo:
    d: dict[int, str]

from_json(Foo, to_json(Foo({1: "a"})))
# SerdeError: ... dict key str '1' not instance of int
```

JSON stringifies object keys, so the key arrives as `"1"` (or `"1.5"` for floats) and the strict type check / float number guard rejects it, even though `from_json(to_json(obj)) == obj` should hold. YAML and `to_dict`/`from_dict` already round-trip these fine, and `dict[IntEnum, V]` keys were recently fixed for JSON too, so plain numeric keys were the odd ones out.

This coerces stringified numeric dict keys back to their declared type when rendering the deserializer, mirroring the existing IntEnum/IntFlag key handling. `bool` is intentionally left out since `bool("false")` is `True`. msgpack/toml are unchanged (they can't represent non-string keys at the format level).

Added a round-trip test covering `dict[int, str]` and `dict[float, str]`; it fails before the change and passes after. Full test suite is green.

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -1160,11 +1160,31 @@ class Renderer:
                 # When the callable type is non generic type e.g int, Foo.
                 callable = v.type.__name__
             return f"collections.defaultdict({callable}, \
-                    {{{self.render(k)}: {self.render(v)} for k, v in {arg.data}.items()}})"
+                    {{{self.dict_key(k)}: {self.render(v)} for k, v in {arg.data}.items()}})"
         else:
             k = arg.key_field()
             v = arg.value_field()
-            return f"{{{self.render(k)}: {self.render(v)} for k, v in {arg.data}.items()}}"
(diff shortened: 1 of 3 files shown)


yukinarit (changes requested):
Hi @sarathfrancis90! 
Thanks for PR. Can you please fix the merge conflict?

sarathfrancis90:
Done — rebased onto main and resolved the conflict. main had since extracted `_deserialize_optional` from `from_obj`, which freed up its complexity budget, so I dropped my earlier `_from_obj_dict` extraction; the change is now just the `dict_key` coercion in the renderer (plus the tests). `make check` and the full suite pass locally.

yukinarit (approved):
LGTM, thanks!

PR #773, opened 2026-07-01
Fix alias for non-primitive fields

### LINKED ISSUE #432: `alias` broken for `list`
Hi there,

Back again with another potential bug. Using `pyserde@v0.12.3`, I've run into deserialization failures for two situations I think should be valid:

```py
@serde
@dataclass
class Foo:
    a: Optional[list[int]] = field(alias=["b"], default=None)

x = from_json(Foo, '{"b": [1]}')
assert x.a == [1] 💥
```

```py
@serde
@dataclass
class Foo:
    a: list[int] = field(alias=["b"], default_factory=list)

x = from_json(Foo, '{"b": [1]}')
assert x.a == [1] 
```

Both fail with `serde.compat.SerdeError: 'a'`, I wish I could give you a more useful error message.

sarathfrancis90:
Deserializing a field with an `alias` only worked when the field was a primitive, enum or `Optional`. For any other type it failed. The two snippets from #432 are the clearest example:

```python
@serde
@dataclass
class Foo:
    a: Optional[list[int]] = field(alias=["b"], default=None)

from_json(Foo, '{"b": [1]}')  # SerdeError: missing required field 'a'
```

The generated deserializer read `data["a"]` directly, so any aliased key was ignored for lists, sets, dicts, tuples, datetimes, `Decimal`, nested dataclasses, etc. A `tuple` alias was worse — the alias leaked into the element fields and produced invalid generated code, so the class couldn't even be defined.

The alias handling was duplicated in a couple of leaf renderers (`primitive`, `enum`) instead of living where the value is actually read. I moved it into `DeField.data`, so every renderer resolves the value through `_get_by_aliases`, and removed the now-redundant per-renderer handling. `Optional` already worked and still does.

Added tests covering the two cases from the issue plus list/set/dict/tuple/datetime and a nested dataclass (the case @kaiaw raised in the thread).

Fixes #432

### DIFF
--- a/serde/de.py
+++ b/serde/de.py
@@ -835,6 +835,9 @@ class DeField(Field[T]):
         e.g. Optional
             * datavar property returns "data"
             * data property returns "data.get("field_name")".
+        e.g. a field with an alias
+            * datavar property returns "data"
+            * data property returns "_get_by_aliases(data, ["field_name", ...])".
         For other types
             * datavar property returns "data"
             * data property returns "data["field_name"]".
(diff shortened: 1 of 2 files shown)


yukinarit (approved):
LGTM

PR #774, opened 2026-07-01
Fix Enum deserialization for heterogeneous enums

### LINKED ISSUE #585: Enum deserialization doesn't work with non-simple enum values

gaoflow:
`deserialize_enum()` samples the enum's value type from only the first member:

```python
member_value_type = type(next(iter(typ.__members__.values())).value)
if isinstance(value, list) and member_value_type is tuple: ...
if isinstance(value, str) and member_value_type is not str: ...
```

Enums may mix value types, and the relevant member is not always first. When it isn't, both JSON/YAML coercion fallbacks misfire and a value that serialized fine fails to deserialize on the round trip:

```python
import enum
from serde.core import deserialize_enum

class Mixed(enum.Enum):
    NUM = 1            # first member: value type sampled as int
    TXT = "b"
    PAIR = ("x", "y")  # tuple-valued member, not first

# to_json(Mixed.PAIR) produces the list ["x", "y"]; deserializing it back:
deserialize_enum(Mixed, ["x", "y"])
# ValueError: ['x', 'y'] is not a valid Mixed   (expected Mixed.PAIR)
```

A numeric member that is not first fails the same way when used as a dict key: it arrives stringified (e.g. `"30"`), but the sampled type is `str`, so no coercion happens.

This is issue #585; the recently merged #765 fixed only the homogeneous case.

The fix derives the coercion from the incoming `value` itself instead of assuming every member shares the first member's value type: a `list` is retried as a `tuple`, and a `str` is retried against the enum's non-`str` member value types. It is purely additive (the new branches run only in the `except` path that previously re-raised), so no existing test changes and the full suite stays green.

Added regression tests following the existing `test_enum_with_tuple_value` convention: a direct `deserialize_enum` helper test plus parametrized json/yaml/dict round-trips for a not-first tuple member and a not-first numeric dict key.

Fixes #585

### DIFF
--- a/serde/core.py
+++ b/serde/core.py
@@ -1157,14 +1157,25 @@ def deserialize_enum(typ: type[enum.Enum], value: Any) -> enum.Enum:
     try:
         return typ(value)
     except (ValueError, KeyError):
-        # A ValueError/KeyError here means ``typ`` has at least one member (an
-        # empty enum would raise TypeError instead), so ``__members__`` is never
-        # empty in this branch.
-        member_value_type = type(next(iter(typ.__members__.values())).value)
-        if isinstance(value, list) and member_value_type is tuple:
-            return typ(tuple(value))
-        if isinstance(value, str) and member_value_type is not str:
-            return typ(member_value_type(value))
+        # JSON/YAML coerce enum values used as object keys to ``str`` and turn
+        # tuple values into ``list``. Retry after undoing that coercion. The enum
+        # may be heterogeneous, so derive the conversion from ``value`` itself
+        # rather than assuming every member shares the first member's value type.
+        if isinstance(value, list):
+            # A ``list`` can only originate from a tuple-valued member: list
+            # values are unhashable and are never valid enum values.
+            try:
+                return typ(tuple(value))
+            except (ValueError, KeyError):
+                pass
+        elif isinstance(value, str):
+            for member_value_type in {type(member.value) for member in typ}:
+                if member_value_type is str:
+                    continue
+                try:
+                    return typ(member_value_type(value))
+                except (ValueError, KeyError, TypeError):
+                    continue
         raise
 
 
--- a/tests/test_basics.py
+++ b/tests/test_basics.py
@@ -339,6 +339,65 @@ def test_enum_with_tuple_value(se: Any, de: Any) -> None:
     assert de(COpt, se(COpt(None))) == COpt(None)
 
 
+def test_deserialize_enum_helper_heterogeneous() -> None:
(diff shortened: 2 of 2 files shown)


yukinarit (approved):
LGTM

PR #778, opened 2026-07-25
fix: serialize bare enum members and enum dict keys

### LINKED ISSUE #468: Be able to serialize dicts with Enum keys to JSON
It would be nice if this worked:

```py
from enum import Enum

from serde.json import to_json


class MyEnum(Enum):
    FOO = 'foo'
    BAR = 'bar'


my_dict: dict[MyEnum, str] = {
    MyEnum.FOO: 'foo',
    MyEnum.BAR: 'bar',
}

print(to_json(my_dict))
```

But currently (`0.13.0`) is:
```
  File "/workspaces/ng/serde_enum_dict.py", line 16, in <module>
    print(to_json(my_dict))
          ^^^^^^^^^^^^^^^^
  File "/opt/python/3.11.6/lib/python3.11/site-packages/serde/json.py", line 70, in to_json
    return se.serialize(to_dict(obj, c=cls, reuse_instances=False, convert_sets=True), **opts)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/python/3.11.6/lib/python3.11/site-packages/serde/json.py", line 45, in serialize
    return json_dumps(obj, **opts)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/python/3.11.6/lib/python3.11/site-packages/serde/json.py", line 19, in json_dumps
    return orjson.dumps(obj, **opts).decode()  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Dict key must be str
```

Sanjays2402:
Closes #468

`to_obj()` has no enum branch, so an `Enum` member reaches the backend serializer unconverted: `to_json(MyEnum.FOO)` and an enum used as a dict key both raise `TypeError` from `json.dumps`. Mapping values were converted recursively but keys were passed through untouched.

Fixed by converting enum instances with the existing `enum_value()` helper and applying the recursive conversion to mapping keys as well as values. New test in `tests/test_se.py` fails without the change and passes with it; full suite 5246 passed.

### DIFF
--- a/serde/se.py
+++ b/serde/se.py
@@ -430,6 +430,8 @@ def to_obj(
         class_serializer = find_global_class_serializer(type(o))
         if class_serializer is not None:
             return class_serializer.serialize(o)
+        if is_enum(type(o)):
+            return enum_value(type(o), o)
         if is_dataclass_without_se(o):
             # Do not automatically implement beartype if dataclass without serde decorator
             # is passed, because it is surprising for users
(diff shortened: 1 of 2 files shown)


yukinarit (approved):
LGTM, thank you!

PR #781, opened 2026-07-27
Add custom coercer API

### LINKED ISSUE #779: Custom coercer API
`coerce` is a very useful feature that allows common conversion from `tuple` to `list` or `int` to `float`. However, it can also bring other problems. For example, allowing `float` to `int` conversion can sometimes introduce confusion, because when a field is defined as an integer, inputs like `3.5` will be silently converted to `3`. Another example is `str` to `bool` which can convert `"false"` to `False`.

What I'm suggesting here is a custom coerce API. For example:

```python
def integer_only(
    owner: str | None, field: str, target: type[Any], value: Any
) -> Any:
    if target is int:
        if isinstance(value, bool) or not isinstance(value, int):
            location = f"{owner}.{field}" if owner else field
            raise TypeError(f"{location} must be an integer")
    return target(value)


@serde(type_check=coerce(coercer=integer_only))
class Config:
    count: int
```

I think this is a relatively small change. I will submit a PR if you are happy with this change.

AllanChain:
Allow `coerce` to use custom rules for rejecting or customizing conversions.

Closes #779

### DIFF
--- a/docs/en/type-check.md
+++ b/docs/en/type-check.md
@@ -69,6 +69,56 @@ foo = Foo(10)
 print(to_json(foo))
 ```
 
+### Custom coercion rules
+
+By default, `coerce` converts a primitive value by calling its declared type, as in `int(value)`.
+To validate values or use a different conversion, pass a coercer function to
+`coerce(coercer=...)`. The function receives:
+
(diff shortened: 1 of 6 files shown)


yukinarit (approved):
LGTM!

PR #783, opened 2026-08-05
fix: unwrap PEP 695 type aliases in iter_literals

### LINKED ISSUE #782: PEP 695 type alias to Literal fails to deserialize as dataclass field
Using a PEP 695 type alias for a `Literal` field fails to deserialize, while the same `Literal` annotation works when written directly.

Reproducer:

```python
from typing import Literal

from serde import serde
from serde.json import from_json


@serde
class Direct:
    value: Literal["trailing", "mean-reversion"]


type Mode = Literal["trailing", "mean-reversion"]


@serde
class Alias:
    value: Mode


print(from_json(Direct, '{"value": "trailing"}'))
print(from_json(Alias, '{"value": "trailing"}'))
```

Actual result with pyserde 0.31.8:

```text
Direct(value='trailing')
  File "<python-input-0>", line 21, in <module>
    print(from_json(Alias, '{"value": "trailing"}'))
          ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/myuan/.cache/uv/archive-v0/YwqBXBZc2eohy-M-/lib/python3.13/site-packages/serde/json.py", line 149, in from_json
    return from_dict(
        c,
    ...<2 lines>...
        deserialize_numbers=deserialize_numbers,
    )
  File "/home/myuan/.cache/uv/archive-v0/YwqBXBZc2eohy-M-/lib/python3.13/site-packages/serde/de.py", line 686, in from_dict
    return from_obj(
        cls,
    ...<3 lines>...
        deserialize_numbers=deserialize_numbers,
    )
  File "/home/myuan/.cache/uv/archive-v0/YwqBXBZc2eohy-M-/lib/python3.13/site-packages/serde/de.py", line 636, in from_obj
    raise SerdeError(e) from None
serde.compat.SerdeError: missing required field 'value' while deserializing Alias
```

Expected result:

```text
Direct(value='trailing')
Alias(value='trailing')
```

This appears to be related to PEP 695 `TypeAliasType` handling in literal discovery. `iter_unions()` already unwraps PEP 695 aliases, but `iter_literals()` does not, so the literal deserializer is not generated for `type Mode = Literal[...]` when used as a dataclass field. It also related to #753 and #757.

I've checked, this commit(https://github.com/yukinarit/pyserde/compare/main...myuanz:pyserde:main) should resolve the issue. I can submit a PR if needed.

Sanjays2402:
Closes #782

`iter_unions` unwraps PEP 695 aliases via `is_pep695_type_alias`, but `iter_literals` does not, so a `type Mode = Literal["a", "b"]` field never had its literal deserializer generated and `from_json` failed with "missing required field". Added the same unwrapping branch, plus a regression test beside the existing PEP 695 alias tests (fails without the fix, passes with it).

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/serde/compat.py
+++ b/serde/compat.py
@@ -466,6 +466,8 @@ def iter_literals(cls: type[Any]) -> list[TypeLike]:
         if is_union(cls):
             for arg in type_args(cls):
                 recursive(arg)
+        elif is_pep695_type_alias(cls):
+            recursive(cls.__value__)
         if is_dataclass(cls):
             stack.append(cls)
             if isinstance(cls, type):
--- a/tests/test_type_alias.py
+++ b/tests/test_type_alias.py
@@ -109,6 +109,26 @@ class Baz:
     assert value == from_dict(BazClass, to_dict(value))
 
 
+@pytest.mark.skipif(sys.version_info < (3, 12), reason=_PEP695_REASON)
+def test_pep695_type_alias_literal() -> None:
+    ns: dict[str, object] = {"serde": serde, "Literal": typing.Literal}
+    exec(
+        """
+type Mode = Literal["trailing", "mean-reversion"]
+
+@serde
+class Foo:
+    value: Mode
+""",
+        ns,
+    )
+    FooClass = typing.cast(type, ns["Foo"])
+
+    value = FooClass("trailing")
+    assert {"value": "trailing"} == to_dict(value)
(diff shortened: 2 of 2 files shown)


yukinarit (approved):
LGTM, thanks!

PR #798, opened 2026-09-19
Fix internal tagging with deny_unknown_fields

AllanChain:
Previously, internal tagging of unions would pass all keys to the dataclass, including the tag, which can cause problems when used in combination with `deny_unknown_fields`. This PR fixes this by skipping the tagging key.

Generated by DeepSeek V4.1 Flash and manually reviewed.


yukinarit (approved):
LGTM

